PR #2, opened 2020-07-09
Sets some base integrations file structure

josejulio:


spinder:
Merging. Helpful setup.

PR #3, opened 2020-07-10
NOTIF-4 Adding Integrations page

josejulio:
- Using fixed data to play with
  - Collapse integration and basic toolbar

Peek 2020-07-10 11-08

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "dependencies": {
     "@babel/runtime": "^7.7.2",
     "@patternfly/patternfly": "4.10.31",
-    "@patternfly/react-core": "4.18.5",
+    "@patternfly/react-core": "4.23.1",
     "@patternfly/react-icons": "4.3.5",
     "@patternfly/react-styles": "4.3.4",
     "@patternfly/react-table": "4.5.7",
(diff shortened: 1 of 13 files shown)


spinder:
I can't get this to compile locally and my ide hates it too. Something is up with this.  
...ERROR @patternfly/react-table/src/components/Table/RowWrapper.tsx
[tsl] ERROR in .../node_modules/@patternfly/react-table/src/components/Table/RowWrapper.tsx(32,15)
      TS2352: Conversion of type 'undefined' to type 'boolean' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
+ I've bumped patterfly/react-table to the latest version and same issue.  Maybe ts version in package.json is wrong? Or some extra 'dev' only components that cause different behavior?  Something too strict in default compilation rules. I tried to debug but couldn't. Giving up for now.

spinder:
hang on .. does I need to rebuild with this version? https://github.com/RedHatInsights/insights-common-typescript/pull/5

josejulio:
Not need to rebuild, use a “yarn install” and retry, else I might have an error there.

spinder:
Yip. All good now. Merging.

PR #5, opened 2020-07-14
Notif 5: Create/Edit UI

spinder:


src/pages/Integrations/List/Page.tsx
```diff
@@ -35,8 +36,12 @@ export const IntegrationsListPage: React.FunctionComponent = () => {
     const integrationRows = useIntegrationRows(integrations);
     const history = useHistory();
 
+    const [activateModal,updateModal] = useState(false);
+    const [model,updateModel] = useState([integrationRows.rows]);
```
josejulio: You probably want to replace the "integrations" variable above with a state, e.g.
```
const [ integrations, setIntegrations ] = useState<Array<Integration>>([
    {
        id: 'foo',
        isEnabled: true,
        name: 'Aha',
        type: IntegrationType.HTTP,
        url: 'https://aha.com'
    },
    {
        id: 'foo-2',
        isEnabled: true,
        name: 'Pager duty',
        type: IntegrationType.HTTP,
        url: 'https://pagerduty.com/weebhook/thatthis'
    }
]);
```
and remove this line, rows will actually read from integration so any change you make it there will reflect on the list.

src/pages/Integrations/List/Page.tsx
```diff
@@ -35,8 +36,12 @@ export const IntegrationsListPage: React.FunctionComponent = () => {
...
+
     const onAddIntegration = React.useCallback(() => {
-        history.push(linkTo.addIntegration());
+        // history.push(linkTo.addIntegration());
+        updateModal(true);
```
josejulio, line 43: I think that you are confusing how the `useState` hook works. What exactly are you trying to achieve here?

josejulio: Nevermind, I was actually confused by the name. (updateModal / updateModel)

src/pages/Integrations/Create/CreatePage.tsx
```diff
@@ -0,0 +1,108 @@
...
+} from '@patternfly/react-core';
+import { useEffect, useState } from 'react';
+import { IntegrationType } from "../../../types/Integration";
+
+const EMPTY = '';
+const title = Messages.components.integrations.toolbar.actions;
+const choose = Messages.common.choose;
+const webhookType = Messages.pages.integrations.types.hooks;
+
+const CreatePage = props => {
```
josejulio: Add some typing here:

```ts
interface CreatePageProps {
  isModalOpen: boolean;
  ...
}

const CreatePage: React.FunctionComponent<CreatePageProps> = props => {
```

src/pages/Integrations/List/Page.tsx
spinder: To activate the Modal within the click action handler...

src/pages/Integrations/Create/CreatePage.tsx
```diff
@@ -0,0 +1,108 @@
...
+    const [pageModal, updatePageModal] = useState({...props.isModalOpen});
```
josejulio: Is not advised to create state out of a prop, as these can change and the argument of `useState` is only an initializer, it won't pick the changes.

src/pages/Integrations/Create/CreatePage.tsx
```diff
@@ -0,0 +1,108 @@
...
+                name: 'Pager duty'+intCount,
+                type: IntegrationType.HTTP,
+                url: newUrl
+            }
+            );
+        }
+    }
+    const handleExit = () => {
+        updatePageModal(!pageModal);
+        props.updateModal(!pageModal);
```
josejulio: The `props.updateModal(!pageModal);` is fine.

I think we can remove the `updatePageModal(!pageModal);` (along with the `const [pageModal, updatePageModal] = useState({...props.isModalOpen});`)

src/pages/Integrations/Create/CreatePage.tsx
spinder: ah.. that's one problem.  I was attempting to reuse the 'state' already created in the parent page... Hm...

josejulio: Is fine to reuse the state. What i mean is that you don't need to create your own state, just use `props.isModalOpen`.

src/pages/Integrations/List/Page.tsx
josejulio: and the `setIntegrations` will be equivalent to your `updateModel`

src/pages/Integrations/Create/CreatePage.tsx
spinder: If not by props... how would you pass in the state from the parent page? Or are you suggesting we just duplicate the state here too?

src/pages/Integrations/Create/CreatePage.tsx
```diff
@@ -0,0 +1,108 @@
...
+    const [pageModal, updatePageModal] = useState({...props.isModalOpen});
+    const [newUrl, updateUrl] = useState("(enter hooks url)");
+    const [intCount,updateIntCount] = useState(0);
+    const handleTextFieldChanges = (evt) => {
+        updateUrl(evt);
+    };
+    const handleSubmit = () => {
+        if(newUrl){
+            updateIntCount(intCount+1);
+            props.updateModel(
```
josejulio: Assuming the model is an array, you should do:

```ts
// the setter accepts either the value or a function that gets the prevValue and returns the new value
// This is better when you are updating/extending the prevModel because it always gets the last state in the case
// of multiple calls to the setter (async calls, other places in the code, etc)
props.updateModel(prevModel => {
   return [
      ...prevModel,
      {
         id: 'newAddition-'+intCount,
         isEnabled: true,
         name: 'Pager duty'+intCount,
         type: IntegrationType.HTTP,
         url: newUrl
      }
   ];
});
```

spinder:
I'd like to add some tests for this.  But functionally it should be complete.

src/Routes.tsx
```diff
@@ -25,11 +26,11 @@ const pathRoutes: Path[] = [
...
-    {
-        path: linkTo.addIntegration(),
-        component: AddNotificationPagePlaceholder,
-        rootClass: 'integrations'
-    },
+    // {
+    //     path: linkTo.addIntegration(),
+    //     component: AddNotificationPagePlaceholder,
+    //     rootClass: 'integrations'
+    // },
```
josejulio: Lets remove this commented code.

src/pages/Integrations/Create/CreatePage.tsx
```diff
@@ -0,0 +1,156 @@
...
+    const [intCount,updateIntCount] = useState(0);
+    const [currentRow,updateCurrentRow] = useState(props.currentRow);
+    const [pageTitle,updatePageTitle] = useState(Messages.components.integrations.toolbar.actions.addIntegration);
+    useEffect(() => {
+        if(props.currentRow){
+            updatePageTitle(Messages.components.integrations.toolbar.actions.editIntegration);
+            let integration = props.model.find(element => element.id == props.currentRow);
+            //preload url for editing if available.
+            if(integration){
+                updateUrl(integration.url);
```
josejulio, line 38: `formik` would fill this (and any other data that we have), it would be good to use it (in a separate PR), as this page will be more complex when more types appears.

josejulio (commented):
I think this is fine to merge by just removing the commented code and fixing the lint issues.

src/pages/Integrations/Create/CreatePage.tsx
spinder: formik can do this too? Hmm.. I'm still reading up on it now. I'll look to address that in a another PR then.

spinder:
I made the requested changes. I forgot to re-enable 'lint'.

josejulio (approved):

PR #22, opened 2020-09-18
NOTIF-7 Notifications table layout & edit forms

josejulio:
Peek 2020-09-22 08-16

Screenshot_2020-09-30_12-51-06
Screenshot_2020-09-30_12-51-20


src/app/App.scss
```diff
@@ -6,3 +6,7 @@
 .pf-c-wizard__main-body {
   max-height: 100%;
 }
+
+.pf-c-page__main#root {
+  --pf-c-page__main-section--BackgroundColor: #FFF;
```
AllenBW: might want to use the pf global var here `var(--pf-global--Color--light-100)` (same value)

src/components/Notifications/ActionComponent.tsx
```diff
@@ -0,0 +1,76 @@
+import * as React from 'react';
+import { BellIcon, EnvelopeIcon, SlackHashIcon } from '@patternfly/react-icons';
```
AllenBW: might want to grab these individually from the /dist/ems (or js) folder means we don't have to bring in whole lib

AllenBW: something like `import BellIcon from '@patternfly/react-icons/dist/ems/icons/bell-icon';`

src/components/Notifications/ActionComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+    isDefault: boolean;
+    action: Action | undefined;
+}
+
+interface ActionTypeToIconProps {
+    actionType: NotificationType;
+}
+
+const marginLeftClassName = style({
+    marginLeft: Spacer.SM
```
AllenBW, line 20: pf constants are always preferred to local ones... alllssooooooo m a yuge fan of separating style from other code (some scss somewhere) but not sure if either of those changes are easily manageable with line and 23

src/components/Notifications/ActionComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    if (props.isDefault) {
+        return (
+            <ActionComponentWrapper { ...props }>
+                <span>Default behavior</span>
+            </ActionComponentWrapper>
+        );
+    }
+
+    if (props.action === undefined) {
```
AllenBW:
```suggestion
    if (!props.action) {
```

src/components/Notifications/ActionComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+        return (
+            <ActionComponentWrapper { ...props }>
+                <span className={ grayFontClassName }>
```
AllenBW, line 60: might want to use the pf text component here... https://patternfly-react.surge.sh/components/text

src/components/Notifications/DefaultBehavior.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { style } from 'typestyle';
+import { ActionComponent } from './ActionComponent';
+import { Button, ButtonVariant, Flex, FlexItem } from '@patternfly/react-core';
+
+export interface DefaultBehaviorProps extends OuiaComponentProps {
+    defaultBehavior: DefaultNotificationBehavior;
+    onEdit: () => void;
+}
+
+const contentClassName = style({
```
AllenBW, line 14: same comments about styling separation n using pf constants as before

src/components/Notifications/DefaultBehavior.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react';
+import { OuiaComponentProps, Spacer } from '@redhat-cloud-services/insights-common-typescript';
+import { NotificationType, DefaultNotificationBehavior } from '../../types/Notification';
+import { getOuiaProps } from '../../utils/getOuiaProps';
+import { style } from 'typestyle';
+import { ActionComponent } from './ActionComponent';
+import { Button, ButtonVariant, Flex, FlexItem } from '@patternfly/react-core';
```
AllenBW, line 7: these can also be imported piecemeal from ah `/dist/ems(umd/js)`

AllenBW, line 7: Flex, ie layouts from `@patternfly/react-core/dist/ems/layouts/`

src/components/Notifications/Form.tsx
```diff
@@ -0,0 +1,178 @@
+import * as React from 'react';
+import {
+    Action,
+    NotificationType,
+    DefaultNotificationBehavior,
+    Notification,
+    IntegrationRef
+} from '../../types/Notification';
+import { Button, ButtonVariant } from '@patternfly/react-core';
```
AllenBW, line 9: same dist comments as above

src/components/Notifications/Form.tsx
```diff
@@ -0,0 +1,178 @@
...
+}
+
+interface ActionsArrayProps extends FieldArrayRenderProps {
+    form: FormikProps<Notification | DefaultNotificationBehavior>;
+    type: Type;
+    getRecipients: (search: string) => Array<string>;
+    getIntegrations: (type: IntegrationType, search: string) => Array<IntegrationRef>;
+}
+
+const alignLeftClassName = style({
```
AllenBW, line 39: same style things as before, constants n seperated

src/components/Notifications/DefaultBehavior.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+});
+
+const titleClassName = style({
+    fontWeight: 600
+});
+
+export const DefaultBehavior: React.FunctionComponent<DefaultBehaviorProps> = (props) => {
+    return (
+        <div { ...getOuiaProps('Notifications/DefaultBehavior', props) } className={ contentClassName } >
```
AllenBW: yah can kill the 
```
{
    return (
```
from above n just return this

src/components/Notifications/SaveModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+    data: Notification;
+}) & DataFetcher
+
+interface InternalProps extends DataFetcher {
+    onClose: (saved: boolean) => void;
+    type: NotificationSaveModalProps['type'];
+}
+
+const InternalNotificationSaveModal: React.FunctionComponent<InternalProps> = (props) => {
+    const title = props.type === 'default' ? 'Edit default notification actions' : 'Edit notification actions';
```
AllenBW:
```suggestion
    const title =  `Edit${props.type === 'default' && ' default'} notification actions;
```

src/components/Notifications/Table.tsx
```diff
@@ -0,0 +1,346 @@
...
+import { GroupByEnum } from './Types';
+
+import './Table.scss';
+
+const pfBorderBottomClassName = style({
+    borderBottom: 'var(--pf-c-table--border-width--base) solid var(--pf-c-table--BorderColor)'
+});
+
+const noBorderBottom = style({
+    borderBottom: 'none !important'
```
AllenBW, line 30: never fun to rely on !important, importing ah .scss directly into this component would likely eliminate the need for this

(or maybe not because yer build pipeline is different? 🤔 )

src/components/Notifications/Table.tsx
```diff
@@ -0,0 +1,346 @@
...
+                        event={ notification.event }
+                    />,
+                    props: {
+                        rowSpan,
+                        className: joinClasses(
+                            noExpandableBorderClassName,
+                            pfBorderBottomClassName
+                        ),
+                        colSpan: parent === undefined ? 1 : 1 // Todo: Change this to `? 1 : 2` once the PF bug is fixed
+                        // https://github.com/patternfly/patternfly-react/issues/4858
```
AllenBW, line 183: ❤️

src/components/Notifications/Table.tsx
```diff
@@ -0,0 +1,346 @@
...
+                }
+            ],
+            props: {
+                className: notification.actions.length > 1 ? noBorderBottom : ''
+            }
+        });
+
+        if (parent !== undefined) {
+            rows[rows.length - 1].parent = parent;
+            // rows[rows.length - 1].useAllCellInExpandedContent = true;
```
AllenBW: comment we can kill?

src/components/Notifications/Table.tsx
```diff
@@ -0,0 +1,346 @@
...
+                    }
+                ],
+                props: {
+                    className: (i + 1 === rowSpan ? '' : noBorderBottom)
+                }
+            });
+
+            if (parent !== undefined) {
+                rows[rows.length - 1].parent = parent;
+                // rows[rows.length - 1].useAllCellInExpandedContent = true;
```
AllenBW: comment we can kill?

src/components/Notifications/Table.tsx
```diff
@@ -0,0 +1,346 @@
...
+                rows[rows.length - 1].fullWidth = true;
+            }
+        }
+
+        return rows;
+    }, [] as Array<IRow>);
+};
+
+const toTableRowsGroupedByApplication = (applicationGroups: Array<NotificationRowGroupedByApplication>, onEdit: OnEditNotification): Array<IRow> => {
+    return applicationGroups.reduce((rows, applicationGroup) => {
```
AllenBW: can kill the `{ return `n  save a line, return the `applicationGroups.` ...

src/pages/Notifications/List/Page.tsx
```diff
@@ -1,27 +1,288 @@
...
+                recipient: [
+                    'Admin'
+                ]
+            },
+            {
+                type: NotificationType.INTEGRATION,
+                integration: {
+                    type: IntegrationType.WEBHOOK,
+                    id: 'integration-00001',
+                    name: 'Send stuff over there'
```
AllenBW, line 96: 😆

package.json
```diff
@@ -4,22 +4,23 @@
...
-    "@redhat-cloud-services/insights-common-typescript": "1.0.5",
+    "@redhat-cloud-services/insights-common-typescript": "3.0.1",
     "@redhat-cloud-services/rbac-client": "1.0.71",
     "axios": "^0.19.2",
     "camelcase": "5.3.1",
     "classnames": "^2.2.6",
     "csstips": "^1.2.0",
     "csx": "^10.0.1",
     "date-fns": "^2.9.0",
+    "emotion": "^10.0.27",
```
AllenBW: professional preference, lockin down all deps to exact versions, less surprises 🤷 🏆 :trollface: 🤗

AllenBW (commented):
HIIIIII this is exciting, gotta lotta question marks about how this all works together that m workin on piecing together... but here are some high level thoughts...

ah few feedbacks got a lil repetitive so i stopped doin em, but they apply everywhere, using distribution versions of deps, styling separate from and consumed by components (rather than baked in) and then then use of pf variables for all styling things (when possible)...

🥇 💯 🍍

src/components/Notifications/Table.tsx
josejulio: The issue I encountered with this syntax is that PF was added *after* these kind of declarations, thus PF definition won here.
I suppose that some way to counter this is to check if we can bump the priority of typestyle.

package.json
josejulio: Actually this shouldn't even be here. Had to do some imports when testing with patternfly/react-core. I'll just remove this.

src/components/Notifications/ActionComponent.tsx
josejulio: I though this was done already (grabbing individual code) but looks like it doesn't... I wonder if there are other ways to avoid the full-syntax per import.

src/components/Notifications/ActionComponent.tsx
josejulio: I prefer to have the code connected when dealing with jsx/tsx. Is easier for me to quickly see where this class is applied and remove stale elements. The linter will usually let me know if this goes unused or if I'm using a wrong value in there.

package.json
josejulio: > lockin down all deps to exact versions

Ah, I never got to finish locking all the dependencies, that's something I have to do.

src/components/Notifications/ActionComponent.tsx
josejulio: Could you elaborate about using that component? I suppose I'm missing something from it.

josejulio:
Merging this to be able to have something to see on the notifications page, some review recommendations need changes on the whole app or common lib.

PR #35, opened 2020-11-16
Update deps, add sitemap (does nothing yet)

AllenBW:
nothing appears to break!🤔💯🐘

### no data, but it all looks workin
Screen Shot 2020-11-16 at 1 43 45 PM
Screen Shot 2020-11-16 at 1 43 56 PM


AllenBW:
Yipes, working on that travis test failure

josejulio (changes requested):
I think we should delete the `package-lock.json` and only update the `yarn.lock` file.

josejulio (approved):

PR #37, opened 2020-11-17
Add intl support and integration table zero state 

AllenBW:
fixes https://issues.redhat.com/browse/NOTIF-31

#### looks like
Screen Shot 2020-11-18 at 10 56 13 AM


AllenBW:
:oof: yeah gotta fix the tests, hold on a sec

build/messages/src/Messages.json
```diff
@@ -0,0 +1,18 @@
...
+  },
+  "integrationsTableEmptyStateBody": {
+    "defaultMessage": "Connect to webhooks and external applications.",
+    "description": "Integrations Empty State body"
+  },
+  "notifications": {
+    "defaultMessage": "Notifications",
+    "description": "Application name"
+  }
+}
```
josejulio, line 18: Is this build file required in the commit?

src/properties/Messages.ts
```diff
@@ -1,10 +1,22 @@
...
+    onError: console.log,
+    locale
+}, cache);
+const intlSettings = { locale };
 
-//Capture some strings we reuse. Possibly use in i18n later?
 const MutableMessages = {
-    appName: 'Notifications',
-    appNameIntegrations: 'Integrations',
+    appName: intlHelper(intl.formatMessage(messages.notifications), intlSettings),
```
josejulio, line 18: Maybe we should remove these `appNames` / `appNameIntegrations` and only use the ones in DefinedMessages. Eventually this file will be gone in that case.

src/app/__tests__/App.test.tsx
```diff
@@ -37,7 +41,7 @@ describe('src/app/App', () => {
         });
         (fetchRBAC as jest.Mock).mockImplementation(() => promise);
         render(
-            <App/>,
+            <IntlProvider locale={ navigator.language } messages={ messages }><App /></IntlProvider>,
```
josejulio, line 44: This is using `AppWrapper` already, do we need the `IntlProvider` here (and the same on the rest of the file)?

build/messages/src/Messages.json
AllenBW: yes

src/properties/Messages.ts
AllenBW: yeah! absolutely

build/messages/src/Messages.json
josejulio: Ok. I think we could check how to integrate the generation of this file in the build process (in a separate PR). 
I suppose that is manually updated, right? At least having a check to validate that is "updated" (again, in a separate PR).

src/app/__tests__/App.test.tsx
AllenBW: yeah we do sadly, wish the wrapper was enough but it seems we also need em on each individual render

src/properties/Messages.ts
AllenBW: we can duplicate the page location through the id or description fields if desired 🤔

build/messages/src/Messages.json
AllenBW: We can change the location if yah want, but its the where the extracted strings are stored, from which the translations.json is built  (the location is where other apps throw it 🤷 )

josejulio: My only concern is if we can end up with an old build Messages.json file. Not updated because dev forgot to run the `translations` command. (IIUC)
It can be reviewed on a followup.

src/properties/Messages.ts
josejulio: You mean the URL? It shouldn't be internationalized anyway, so we might put that on the `Config`. Or in what context/file are you referring ?

src/app/__tests__/App.test.tsx
josejulio: :cry:

build/messages/src/Messages.json
AllenBW: yeah that will take a lil diligence, running translate whenever yah add new strings, we could hijack start to also do it... but the string adding isn't always a thing done, ALSO until we're seriously seriously using intl, kinda doesn't quite matter if they get outta date 🤷

src/properties/Messages.ts
AllenBW: oooo was referring to the helpfulness of the `MutableMessages` json in Messages.ts, an example of an id could be... integrations-list-title

PR #38, opened 2020-11-17
NOTIF-50 Hiding some elements from prod-beta & staging-beta

josejulio:
- Filters on Notifications & Integrations
 - Drawer, Notification and Email types
 - Export on Notifications

Marking as Draft because the required features are only on ci-beta and qa-beta (It will break in other environments)


theute:
So we have to wait Nov 30th for Prod beta ?
Any idea about Stage beta ?

josejulio:
> So we have to wait Nov 30th for Prod beta ?

I got told that all environments will have this change by Nov 30th.
I suppose this includes all up to Prod-stable. I'll confirm.

PR #62, opened 2021-01-13
More missaligment fixes

josejulio:
Before:
Screenshot_2021-01-13_14-24-02

After:
Screenshot_2021-01-13_14-02-20


katierik:
Looks great

PR #81, opened 2021-02-17
Migrate app to chrome 2

Hyperkid123:
cc @josejulio 

I have found some issues that have to be fixed in order for the app to work in chrome 2

- the react-router dom package sharing was not working; don't know why yet, I am suspicious it might be caused by the TS build; for now, it's removed from dependency sharing
- app entry was not exporting the app as a `default` module; This is required because of the React.lazy API
- Also I have changed the dev server to use from HTTP instead of HTTPS protocol; Required because of the proxy from /beta to / route for webpack dev server (it was unintentionally working in old chrome because the history API fallback was routing static assets from /beta to / but there are now XHR requests that are not caught by the fallback)

Have to be merged with: https://github.com/RedHatInsights/cloud-services-config/pull/407

### DIFF
--- a/config/dev.webpack.config.ts
+++ b/config/dev.webpack.config.ts
@@ -7,7 +7,7 @@ import { updateTsLoaderRule } from './common.webpack.config';
 const { config: webpackConfig, plugins } = config({
     rootFolder: resolve(__dirname, '../'),
     debug: true,
-    https: true,
+    https: false,
     useFileHash: false
 });
 
@@ -17,7 +17,9 @@ plugins.push(
     federatedModules(
         {
             root: resolve(__dirname, '../'),
-            useFileHash: false
+            debug: true,
+            useFileHash: false,
+            exclude: [ 'react-router-dom' ]
         }
     )
 );
--- a/config/prod.webpack.config.ts
+++ b/config/prod.webpack.config.ts
@@ -12,7 +12,8 @@ plugins.push(
     federatedModules(
         {
             root: resolve(__dirname, '../'),
-            useFileHash: false
+            useFileHash: true,
+            exclude: [ 'react-router-dom' ]
         }
     )
 );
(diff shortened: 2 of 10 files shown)


config/dev.webpack.config.ts
```diff
@@ -7,7 +7,7 @@ import { updateTsLoaderRule } from './common.webpack.config';
 const { config: webpackConfig, plugins } = config({
     rootFolder: resolve(__dirname, '../'),
     debug: true,
-    https: true,
+    https: false,
```
josejulio, line 10: This was part of a test? or is usually better to run in http?

config/prod.webpack.config.ts
```diff
@@ -12,7 +12,8 @@ plugins.push(
     federatedModules(
         {
             root: resolve(__dirname, '../'),
-            useFileHash: false
+            useFileHash: true,
```
josejulio, line 15: Should this be `false` on prod ?

josejulio: Ah, nevermind. True on prod, false on dev. :yawning_face:

config/dev.webpack.config.ts
Hyperkid123: No this is required because we have to proxy the dev server from /beta to / in because the historyAPIfallback no does not work on HXR when trying to request the fed-mods.json file. And because of the invalid certificates, we can't use HTTPS without turning off the security checks.

config/prod.webpack.config.ts
Hyperkid123: No, we want it true on prod to prevent stale assets to be served. We had a lot of issues with this in the past. It also can reveal invalid builds when beta assets are served on stable env and vice versa.

josejulio (approved):

PR #113, opened 2021-05-14
First round at supporting sending via Camel

pilhuhn:
This is a WIP, I don't know what I am doing here. And due to my node-sass issues, I can't even tell if I would see anything on screeen.
This uses the OpenAPI from https://github.com/RedHatInsights/notifications-backend/pull/237


src/types/Integration.ts
```diff
@@ -25,11 +27,20 @@ export interface IntegrationHttp extends IntegrationBase {
     method: Schemas.HttpType;
 }
 
+export interface IntegrationCamel extends IntegrationBase {
+    type: IntegrationType.CAMEL;
+    url: string;
+    sslVerificationEnabled: boolean;
+    secretToken?: string;
+    basicAuth?: string; // TODO this is a map { user, pass}
+    extras?: string; // TODO this is a freeform map<string,string>
```
josejulio: more or less

```suggestion
    basicAuth?: {
        user: string;
        pass: string;
    };
    extras?: Record<string, string>;
```

src/generated/OpenapiIntegrations.ts
```diff
@@ -31,32 +31,6 @@ export namespace Schemas {
     username?: string | undefined | null;
   };
 
-  export const BehaviorGroup = zodSchemaBehaviorGroup();
```
josejulio: These are hidden from the openapi in current's `notification-backend`. I left them here because I'm using it for the behavior group work.

I'll probably move them to other file to allow the code to compile and play nice with the schema generator.

josejulio:
Rebased, updated schema and improved (I hope) some types

josejulio (approved):

PR #139, opened 2021-06-30
Fix mobile styling of 'Create new behavior group' modal 

epwinchell:
Jira: https://issues.redhat.com/browse/RHCLOUD-14023

Old
Screen Shot 2021-06-30 at 12 09 11 PM

New
Screen Shot 2021-06-30 at 12 06 48 PM

### DIFF
--- a/src/components/Notifications/BehaviorGroup/BehaviorGroupForm.tsx
+++ b/src/components/Notifications/BehaviorGroup/BehaviorGroupForm.tsx
@@ -69,14 +69,16 @@ const ActionRow: React.FunctionComponent<ActionRowProps> = props => {
 
     return (
         <>
-            <GridItem span={ 6 }>
+            <GridItem sm={ 12 } md={ 6 }>
+                <b>Actions</b>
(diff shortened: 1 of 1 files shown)


josejulio (changes requested):
Thanks @epwinchell 

One issue with this approach is that we have the header (actions/recipient) each time for each row. I'm not sure how the mobile mock is, but the desktop one should have the header only once.

Screenshot_2021-06-30_11-17-51
Screenshot_2021-06-30_11-16-50

epwinchell:
@josejulio The desktop layout hasn't changed:

Screen Shot 2021-06-30 at 12 21 51 PM

epwinchell:
> One issue with this approach is that we have the header (actions/recipient) each time for each row

@josejulio Good catch. I could use some help with the logic to fix that when someone has time.

josejulio:
> could use some help with the logic to fix that when someone has time.

I could help with that. Do you know how the UI should look in mobiles?

PR #141, opened 2021-07-05
NOTIF-242 Fetch system Endpoints instead of creating

josejulio:
Instead of creating/deleting the Email endpoints, makes use of the new API to fetch the system endpoints to assign them to behavior groups.

Requires https://github.com/RedHatInsights/notifications-backend/pull/362


aferd (approved):
lgtm

PR #143, opened 2021-07-30
NOTIF-255 Adding tooltip to Create group button

josejulio:
Screenshot_2021-07-30_17-46-35


aferd (dismissed):

josejulio:
> failing test

Upgrading patternfly dependencies is almost the same as "Some tests are gonna fail!" :grin:

PR #168, opened 2021-10-19
NOTIF-353 Limit and sort the history list

josejulio:


src/pages/Integrations/List/useIntegrationRows.ts
```diff
@@ -76,11 +76,15 @@ export const useIntegrationRows = (integrations: Array<UserIntegration>) => {
                 limit.clearQueue();
 
                 integrations.map(integration => integration.id).forEach(integrationId => {
-                    limit(() => query(listIntegrationHistoryActionCreator(integrationId))).then(response => {
+                    limit(() => query(listIntegrationHistoryActionCreator({
+                        integrationId,
+                        limit: 5,
+                        sortBy: 'invocation_time:desc'
```
josejulio: @gwenneg Is this the expected sorting? or it should be `created`?

gwenneg: In the backend we only have the `ctime` field (which contains the `Fire time (since the epoch)`) for history items. Do you really have `invocation_time` and `created` available in the frontend?

josejulio: It's not for showing, so not really an issue if we don't have it available on the frontend. 
I didn't found the ctime on the [model](https://github.com/RedHatInsights/notifications-backend/blob/6271e4da707a7b94b8e3bd52d91fb11a9d260aa4/backend/src/main/java/com/redhat/cloud/notifications/models/NotificationHistory.java#L34) so maybe I'm looking a different model?

gwenneg: Oops I was looking at the policiies history earlier and mixed everything 😄

`invocation_time` will contain a duration, so I wouldn't sort on that. I think `created:desc` is the good way to sort and retrieve the 5 last attempts.

josejulio: We will have to update the backend code for that. I'm currently getting: `column reference \"created\" is ambiguous` in the logs when trying that request.

gwenneg: Ok I'll do it in the existing backend PR.

gwenneg: Could you try sending `nh.created:desc` instead of `created.desc`?

josejulio: That works, should we use it that way? maybe something we would have to document on the API for the sortby?

gwenneg: The REST API is tightly coupled with the Hibernate layer because of the `Query` class. That's the reason why I never use that class in new APIs. I think we should use `nh.created:desc` without documenting it for now and get rid of the layers coupling in a subsequent PR. Pushing the limit to prod should be done ASAP because of the prod incident we had recently.

gwenneg:
LGTM

aferd (approved):
lgtm

PR #170, opened 2021-10-29
NOTIF-362 Downgrade react-redux dependency

josejulio:


aferd (approved):
/lgtm

Curious as to why the downgrade was needed, if you have time to chat about it. Thanks!

PR #181, opened 2021-11-23
Prevent from breaking if properties are null

josejulio:
Using [nullish coalescing operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing)


aferd (approved):
lgtm

PR #183, opened 2021-11-25
NOTIF-104 Save actions in the same order they are specified

josejulio:


src/pages/Notifications/Form/useSaveBehaviorGroup.tsx
```diff
@@ -0,0 +1,215 @@
...
+                    operation: SaveBehaviorGroupResult.CREATE,
+                    status: true
+                };
+            } else {
+                return {
+                    operation: SaveBehaviorGroupResult.UPDATE,
+                    status: true
+                };
+            }
+        } else if (behaviorGroupId === undefined) {
```
aferd, line 156: Why is this in here back to back? See lines below (@ 169)

aferd: nvm, I see the return status is different.

josejulio: It's a bit difficult to see with the `if` nesting, but line 169 is outside of the `if actionsNeedsUpdate(original, updated)` at line 105.

So line 169 deals with the case when we didn't need to update, but are returning an OK response anyway.

aferd (approved):
lgtm

PR #195, opened 2021-12-21
NOTIF-429 Uses camel:splunk types

josejulio:
Peek 2021-12-20 12-48

Peek 2022-01-20 10-54

Screenshot_2022-01-20_12-32-45

"AnyCamel" is just a madeup camel integration to test that it won't be really hard to add more types

Instead of using `Camel` as a type, use it's subtype (`Splunk`).
~There are some places where this isn't yet changed (EventLog is one place, there might be others)~

### DIFF
--- a/src/types/adapters/IntegrationAdapter.ts
+++ b/src/types/adapters/IntegrationAdapter.ts
@@ -2,9 +2,11 @@ import { assertNever } from 'assert-never';
 
 import { Schemas } from '../../generated/OpenapiIntegrations';
 import {
+    CamelIntegrationType,
     Integration,
     IntegrationBase,
-    IntegrationCamel, IntegrationEmailSubscription,
+    IntegrationCamel,
+    IntegrationEmailSubscription,
     IntegrationHttp,
     IntegrationType,
     NewIntegration,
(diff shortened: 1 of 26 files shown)


josejulio:
This PR is ready for review, but requires https://github.com/RedHatInsights/notifications-backend/pull/923
Leaving status as draft for now

src/types/adapters/IntegrationAdapter.ts
```diff
@@ -2,43 +2,39 @@ import { assertNever } from 'assert-never';
...
-            return Schemas.EndpointType.Enum.webhook;
-        case IntegrationType.CAMEL:
-            return Schemas.EndpointType.Enum.camel;
-        case IntegrationType.EMAIL_SUBSCRIPTION:
-            return Schemas.EndpointType.Enum.email_subscription;
-        default:
-            assertNever(type);
-    }
+export const getEndpointType = (type: IntegrationType): [ Schemas.EndpointType, string] => {
+    return type.split(':', 2) as [ Schemas.EndpointType, string ];
```
vkrizan: Suggestion: Instead of a custom data structure holding two values, can it be instead split into two separate variables? Either by passing them separately or passing an object (`endpoint_type: {type: String, sub_type: String}`)?

src/types/adapters/IntegrationAdapter.ts
```diff
@@ -120,13 +113,32 @@ export const toIntegration = (serverIntegration: ServerIntegrationResponse): Int
...
-    switch (integration.type) {
+
+    const type = integration.type;
+
+    if (isCamelType(type)) {
+        const integrationCamel: IntegrationCamel = integration as IntegrationCamel;
+        return {
+            url: integrationCamel.url,
+            disable_ssl_verification: !integrationCamel.sslVerificationEnabled,
+            secret_token: integrationCamel.secretToken,
```
vkrizan, line 136: Is a `secret_token` a new parameter for a integration? So far we've used `extras[token]` to pass the token for our [Splunk integration](https://github.com/RedHatInsights/eventing-integrations/blob/efea5f7cd5e253739a19ccf206f52063f0fd4da2/splunk-quarkus/src/main/java/com/redhat/console/notifications/splunkintegration/SplunkIntegration.java#L151).
cc @pilhuhn

vkrizan (commented):
Thank you @josejulio! 26 changes, wow.

Looking forward to this.

aferd (approved):
lgtm

src/types/adapters/IntegrationAdapter.ts
josejulio: I don't think it's new.

src/types/adapters/IntegrationAdapter.ts
josejulio: Sure. Actually this is an internal method (will remove the `export`)

vkrizan: Where does the `type` value come from? Is that also internal?

josejulio: Yes, it's internal to the UI code: see: https://github.com/RedHatInsights/notifications-frontend/blob/master/src/types/Integration.ts#L6-L10
It converts that type to a type/subType that is consumed by the notifications-backend

vkrizan: Oh, so the origin is the constant? If so, then the constant might have it as a struct.
Type-wise something like this, changing the enum into union of several types.
```
export const SPLUNK = Object.freeze({ type: 'camel', sub_type: 'splunk'  });
type IntegrationType = | WEBHOOK | EMAIL_SUBSCRIPTION | SPLUNK;
```

josejulio: That looks nice! I'll play with that idea! Thanks!

vkrizan: No problem. Cheers.

josejulio: Played with this a bit, but ran into some issues where a enum is far easier to manage. (narrowing was one of these issues, as the enum is used with some cases/assertNevers) - I might revisit it again to have a 1-1 match between the server/client.

vkrizan: No worries. My first thought was whether the enum can hold an object, which it probably cannot. For a listing possible types there would have to be additional array (if TS does not allow unwrapping the type definition).

PR #196, opened 2021-12-23
NOTIF-253 BG edit kebab is disabled when user is not OrgAdmin

aferd:


src/components/Notifications/BehaviorGroup/BehaviorGroupCard.tsx
```diff
@@ -61,7 +63,7 @@ const BehaviorGroupCardLayout: React.FunctionComponent<BehaviorGroupCardLayout>
                 <CardActions>
                     <Dropdown
                         onSelect={ switchOpen }
-                        toggle={ <KebabToggle onToggle={ setOpen } isDisabled={ !props.dropdownItems } /> }
+                        toggle={ <KebabToggle onToggle={ setOpen } isDisabled={ !props.dropdownItems || !isOrgAdmin } /> }
```
josejulio: The JIRA mentions `org admin`, but i think we can use the [canWriteNotifications](https://github.com/aferd/notifications-frontend/blob/de7869b03255f820bd8eaea4ea84fb254e5b2995/src/app/AppContext.tsx#L10) property instead.
That way, all users who were given this permission could access this dropdwon.

josejulio (approved):
lgtm

PR #201, opened 2022-01-07
 NOTIF-368 add proper edit/create title when editing behavior group

aferd:


aferd:
https://user-images.githubusercontent.com/86322239/148595861-1d210eb2-e063-4c36-b111-7b73fd6c13bc.mov

src/components/Notifications/BehaviorGroup/BehaviorGroupForm.tsx
```diff
@@ -21,7 +21,8 @@ export const EditBehaviorGroupForm: React.FunctionComponent<EditBehaviorGroupPro
 
     return (
         <div { ... getOuiaProps('Notifications/BehaviorGroupForm', props) }>
-            <div className={ subtitleClassName }>Enter a name and add actions for your new group.</div>
+            <div className={ subtitleClassName }>{`${ props.behaviorGroup?.actions ?
```
josejulio: Could you check for the `id` instead of the actions?

src/components/Notifications/BehaviorGroup/BehaviorGroupSaveModal.tsx
```diff
@@ -19,7 +19,7 @@ interface InternalProps {
 }
 
 const InternalBehaviorGroupSaveModal: React.FunctionComponent<InternalProps> = props => {
-    const title =  `${props.data ? 'Create new' : 'Edit'} behavior group`;
+    const title =  `${props.data?.actions ? 'Edit' : 'Create new'} behavior group`;
```
josejulio: Could you check for the `id` instead of the actions?

It should be more stable - I'm wondering what would happen if you
1- Click "Create new group"
2- Add an action

josejulio (approved):
lgtm

src/components/Notifications/BehaviorGroup/BehaviorGroupSaveModal.tsx
aferd: hm, nothing seemed to change here. I didn't see/get any error when adding actions, but changed to check id!

josejulio: Ah ok - anyhow the `id` should be more future proof. thanks!

PR #228, opened 2022-02-11
NOTIF-286 add text in UI to specify how token in integration is being used.

aferd:
Screen Shot 2022-02-11 at 2 42 41 PM


josejulio (approved):
Perfect!

PR #234, opened 2022-02-16
Adding empty state on settings> notifications

aferd:
Users with readPermissions will not longer be able to see the settings> notifications page.


src/app/App.tsx
```diff
@@ -114,7 +124,14 @@ const App: React.ComponentType = () => {
                     <Routes />
                 </>
             ) : (
-                <NotAuthorized serviceName={ serviceName } />
+                <>
+                    <NotAuthorized
+                        description={ `Contact your organization administrator for more information or visit
+                        ${ myUserAccess } to learn more about your permissions. To manage your notifications,
+                        go to your ${ userPreferences }.` }
```
josejulio: You are transforming the links to string, you want something similar to:

```suggestion
                        description={ <> 
                        Contact your organization administrator for more information or visit
                        { myUserAccess } to learn more about your permissions. To manage your notifications,
                        go to your { userPreferences }.
                        </> }
```

src/app/App.tsx
```diff
@@ -27,12 +28,21 @@ const switchClassname = style({
     padding: 8
 });
 
-const App: React.ComponentType = () => {
+interface GetBundleProps {
+    bundles?: Facet;
```
josejulio: Where is this prop being sent to this componet?

src/app/App.tsx
```diff
@@ -45,13 +55,13 @@ const App: React.ComponentType = () => {
         }
     }, [ intl, location.pathname ]);
 
-    const hasReadPermissions = React.useMemo(() => {
+    const hasWritePermissions = React.useMemo(() => {
```
josejulio: I wonder if instead of changing it to write permission we should remove the read permissions from regular users.
This would be setting this to false: https://github.com/RedHatInsights/rbac-config/blob/master/configs/stage/roles/notifications.json#L22

We can do both stage and prod.

aferd: Okay. Then I would just need to have `rbac ?` as the condition?

src/app/App.tsx
aferd: the link for user pref and my user access.

josejulio: Yes, but I don't see any change when this component (`App` )is being sent this one. I'm guessing it will be `undefined` always.

I think you should probably have to fetch it from the URL

src/app/App.tsx
josejulio: I think the condition would be unchanged (i.e. `hasReadPermissions` would still be relevant)

aferd: ah ok yes, i see now.

aferd:
this is how the page will look once rbac-config has been updated.
https://user-images.githubusercontent.com/86322239/154530761-7c5199a0-9615-47a2-8963-ea3a0b59c6e3.mov

src/app/App.tsx
```diff
@@ -28,12 +29,20 @@ const switchClassname = style({
...
 const App: React.ComponentType = () => {
+
     const intl = useIntl();
     const { rbac, server, isOrgAdmin } = useApp();
     const location = useLocation();
     const insights = getInsights();
     const [ usingExperimental, setUsingExperimental ] = React.useState<boolean>(false);
 
+    const url = window.location.pathname;
+    const filename = url.substring(url.lastIndexOf('/') + 1);
```
josejulio: This is somewhat brittle, but can work for now - The problem we have is that there is no way to know if this result is actually a bundle  (because we have no access to notifications) - We can follow up on this if we see some issues.
Could you update instead of filename call it `bundle` ?

src/app/App.tsx
```diff
@@ -28,12 +29,20 @@ const switchClassname = style({
...
+    const userPreferences = <a href={ localUrl(`/user-preferences/notifications/${ filename }`,
+        getInsights().chrome.isBeta()) }> User Preferences </a>;
+    const myUserAccess = <a href={ localUrl(`/settings/my-user-access?bundle=${ filename }`,
```
josejulio: I was checking, and I think we can fix this `bundle=` to always be `rhel`... Notifications permissions are part of the RHEL bundle.

src/app/App.tsx
aferd: Yeah, i understand that. I can start working towards what you mentioned in slack... because I also need to add the bundle displayName to the page header, so as you said, it would make more sense to split it since we have access to the bundle there.

josejulio: Something else - If the `serviceName` is `integrations`, the `bundle` won't make sense. I think we can set it to an empty string and default to `/user-preferences/notifications/` in the userPreferences.

We could probably have a different message in there, but its fine for now.

aferd:
Currently not switching as intended and still need to display the `bundleDisplayName` vs `bundleName` in the title. Wanted to ask you to look this over to ensure I am headed in the right direction with what you asked.

src/Routes.tsx
```diff
@@ -38,7 +38,7 @@ const pathRoutes: Path[] = [
     },
     {
         path: linkTo.notifications(':bundleName'),
-        component: NotificationsListPage
+        component: NotificationsPermissionsPage
```
josejulio: We are going to need something similar for the other pages

src/components/Notifications/NotificationsPermissions.tsx
```diff
@@ -0,0 +1,67 @@
...
+        const appId = getSubApp(location.pathname);
+        switch (appId) {
+            case Config.integrations.subAppId:
+                return rbac?.canReadIntegrationsEndpoints;
+            case Config.notifications.subAppId:
+                return rbac?.canReadNotifications;
+        }
+
+        return false;
+    }, [ rbac ]);
```
josejulio: This could probably be on a file of its own and be used elsewhere

aferd:
https://user-images.githubusercontent.com/86322239/155618399-4f723da0-be55-4399-b751-ecda67b5530b.mov

aferd:
https://user-images.githubusercontent.com/86322239/155744843-dc1d4989-078d-4882-ace4-5588a899ba84.mov

aferd:
tests are failing in App.test and Routes.test, I may need help re-working those.

src/components/CheckReadPermissions.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+        return false;
+    }, [ rbac ]);
+
+    return (
+        <>
+            { !hasReadPermissions ?
+                (<NotAuthorizedPage />) :
+                <><NotificationsListPage />
+                    <ConnectedIntegrationsListPage /></>
```
josejulio: I don't think this is going to work as you suppose. 
By the looks of this, It will show both applications in the same page (in both /integrations and /notifications). 
I'm going to drop some suggestions

src/components/CheckReadPermissions.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+            }
+        </>
```
josejulio:
```suggestion
        <>
            { !hasReadPermissions ? <NotAuthorizedPage /> : { props.children} }
        </>
```

src/components/NotAuthorized.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export const NotAuthorizedPage: React.FunctionComponent = () => {
+
+    const { bundleName } = useParams<NotificationListPageParams>();
+    const getBundles = useGetBundles();
+    const bundles: Facet | undefined = useMemo (() => {
+        if (getBundles.payload?.status === 200) {
+            return getBundles.payload.value.find(b => b.name === bundleName);
+        }
+    }, [ bundleName, getBundles.payload?.status, getBundles.payload?.value ]);
```
josejulio, line 27: This is not yet going to work - but is something we can update.
IIRC facets require read permissions - we could update it so that it does not require any permission.

src/Routes.tsx
```diff
@@ -30,15 +29,15 @@ const pathRoutes: Path[] = [
     },
     {
         path: linkTo.integrations(),
-        component: ConnectedIntegrationsListPage
+        component: CheckReadPermissions
```
josejulio:
```suggestion
        component: ConnectedIntegrationsListPage
```

src/Routes.tsx
```diff
@@ -30,15 +29,15 @@ const pathRoutes: Path[] = [
...
     },
     {
         path: linkTo.eventLog(),
         component: EventLogPage
     },
     {
         path: linkTo.notifications(':bundleName'),
-        component: NotificationsListPage
+        component: CheckReadPermissions
```
josejulio:
```suggestion
        component: NotificationsListPage
```

src/app/App.tsx
```diff
@@ -89,33 +60,30 @@ const App: React.ComponentType = () => {
...
+                            className={ switchClassname }
+                            isChecked={ usingExperimental }
+                            onChange={ toggleExperimental }
+                            labelOff="Enable experimental features"
+                            label="Disable experimental features"
+                        />
+                    </RenderIfTrue>
+                </InsightsEnvDetector>
+                <Routes />
+            </>
```
josejulio, line 85: Wrapping this all in the `CheckReadPermissions` 
`CheckReadPermissions` will decide if it shows the content or not

```suggestion
            <CheckReadPermissions>
                <NotificationsPortal />
                <InsightsEnvDetector insights={ insights } onEnvironment={ staging }>
                    <RenderIfTrue>
                        <Switch
                            className={ switchClassname }
                            isChecked={ usingExperimental }
                            onChange={ toggleExperimental }
                            labelOff="Enable experimental features"
                            label="Disable experimental features"
                        />
                    </RenderIfTrue>
                </InsightsEnvDetector>
                <Routes />
            <CheckReadPermissions />
```

josejulio (changes requested):

src/components/NotAuthorized.tsx
aferd: ok, yeah, after those fixes this no longer is showing. Would I be looking at the generated files for this change?

josejulio: nevermind my comment above - this api doesn't require any permission

aferd: hmmm.. its coming up undefined now, where it wasn't before.

josejulio: This is because we don't have a router in there -  we can't use the `useParams` as there is nothing to match against.

aferd: yep, spoke too soon, just looked through and saw that.

src/components/NotAuthorized.tsx
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }, [ intl, location.pathname ]);
+
+    const pageHeaderTitleProps = {
+        paddingBottom: '8px'
+    };
+
+    const title = React.useMemo(() => {
+        if (serviceName === 'Notifications') {
+            return `${ serviceName } | ${ bundles?.displayName }`;
```
josejulio: `bundles?.displayName` will be undefined when it's loading. use a <Skeleton> in that case.

src/app/App.tsx
```diff
@@ -74,7 +46,7 @@ const App: React.ComponentType = () => {
         );
     }
 
-    if (server.status === ServerStatus.MAINTENANCE) {
+    {if (server.status === ServerStatus.MAINTENANCE) {
```
josejulio: I think this can be reverted (the curly braces)

src/components/NotAuthorized.tsx
aferd: use a what?

josejulio: sorry - a Skeleton

src/components/CheckReadPermissions.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { useLocation } from 'react-router-dom';
+
+import { useApp } from '../app/useApp';
+import Config from '../config/Config';
+import { getSubApp } from '../utils/Basename';
+import { NotAuthorizedPage } from './NotAuthorized';
+
+export const CheckReadPermissions: React.FunctionComponent = (props) => {
+    const { rbac } = useApp();
```
josejulio: Change to `useAppContext` instead.

`useApp` calls the server again while `useAppContext` yields the already computed value.

src/__tests__/Routes.test.tsx
```diff
@@ -41,7 +41,7 @@ describe('src/Routes', () => {
             });
 
             expect(getLocation().pathname).toBe('/');
-            expect(document.body.firstChild).toBeEmptyDOMElement();
```
aferd, line 44: this test is now showing the NotAuth page, would it make sense to adjust the test to show that or is this still meant to render nothing "/" (ignore the current change in there)

josejulio: I think the no access is fine

aferd: Could you help re-work this test? It is failing because it shows the NotAuth page when running the test

src/app/__tests__/App.test.tsx
```diff
@@ -145,7 +145,7 @@ describe('src/app/App', () => {
             await jest.advanceTimersToNextTimer();
         });
 
-        expect(screen.getByText(/You do not have access to Notifications/i)).toBeTruthy();
+        expect(screen.getByTestId('content')).toBeTruthy();
```
josejulio: Almost there!

Before this change, we were checking the read permissions before the routes. This is no longer true, as the check is inside the routes now - so we need to tweak it to really test the read permissions.

To fix this, we could mock the [ConnectedIntegrationsListPage](https://github.com/RedHatInsights/notifications-frontend/blob/master/src/Routes.tsx#L33) and [NotificationsListPage](https://github.com/RedHatInsights/notifications-frontend/blob/master/src/Routes.tsx#L41) components.

We no longer need to mock the [routes](https://github.com/RedHatInsights/notifications-frontend/blob/da8e0a816af631eaf147b3a22f31e1ba7c500fea/src/app/__tests__/App.test.tsx#L21-L26)

```
jest.mock('../../Routes', () => { // <--- this is the import path to what you want to mock
    const MockedRoutes: React.FunctionComponent = () => <div data-testid="content" />; // something to mock
    return {
        Routes: MockedRoutes // "Routes" is what is being exported, in the case of ConnectedIntegrationsListPage it would be `ConnectedIntegrationsListPage` because that's the name of what we export 
    };
});
```

So we would need a similar block as above for [ConnectedIntegrationsListPage](https://github.com/RedHatInsights/notifications-frontend/blob/master/src/Routes.tsx#L33) and [NotificationsListPage](https://github.com/RedHatInsights/notifications-frontend/blob/master/src/Routes.tsx#L41).

aferd: Those components are already mocked in that file..

```
`jest.mock('../pages/Notifications/List/Page', () => ({
    NotificationsListPage: () => 'Notifications'
}));

jest.mock('../pages/Integrations/List/Page', () => ({
    ConnectedIntegrationsListPage: () => 'Integrations'
}));`
```

Would the above be to replace that or in addition to?

josejulio: Would replace the Routes, also you can wrap the 'Notifications' and 'Integrations' text in a: `<div data-testid="content" />` to prevent from affecting other tests

josejulio (approved):
lgtm

PR #235, opened 2022-02-17
Include actions in eventlog

josejulio:
A new param `includeActions` was added to prevent from collecting the actions when it's not needed. In our case, we need the actions.


aferd:
failed tests, looks like `basicAuth?` is the culprit.

josejulio:
> failed tests, looks like `basicAuth?` is the culprit.

thanks, fixed !

aferd (approved):

PR #244, opened 2022-03-08
adding pr_check and build_deploy

gburges:


.rhcicd/build_deploy.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+# --------------------------------------------
+# Export vars for helper scripts to use
+# --------------------------------------------
+# name of app-sre "application" folder this component lives in; needs to match for quay
+export COMPONENT="notifications"
+export WORKSPACE=${WORKSPACE:-$APP_ROOT} # if running in jenkins, use the build's workspace
+export APP_ROOT=$(pwd)
+export NODE_BUILD_VERSION=12
```
josejulio: I don't think we really have an specific version, but could we use the 16?
```suggestion
export NODE_BUILD_VERSION=16
```

.rhcicd/pr_check.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+# --------------------------------------------
+# Export vars for helper scripts to use
+# --------------------------------------------
+# name of app-sre "application" folder this component lives in; needs to match for quay
+export COMPONENT="notifications"
+export WORKSPACE=${WORKSPACE:-$APP_ROOT} # if running in jenkins, use the build's workspace
+export APP_ROOT=$(pwd)
+export NODE_BUILD_VERSION=12
```
josejulio: same as above
```suggestion
export NODE_BUILD_VERSION=16
```

josejulio (approved):

.rhcicd/build_deploy.sh
```diff
@@ -0,0 +1,34 @@
...
+source <(curl -sSL $COMMON_BUILDER/src/frontend-build.sh)
+BUILD_RESULTS=$?
+
+# Stubbed out for now
+mkdir -p $WORKSPACE/artifacts
+cat << EOF > $WORKSPACE/artifacts/junit-dummy.xml
+<testsuite tests="1">
+    <testcase classname="dummy" name="dummytest"/>
+</testsuite>
+EOF
```
gwenneg, line 31: Is this needed for the frontend?

.rhcicd/pr_check.sh
```diff
@@ -0,0 +1,34 @@
...
+source <(curl -sSL $COMMON_BUILDER/src/frontend-build.sh)
+BUILD_RESULTS=$?
+
+# Stubbed out for now
+mkdir -p $WORKSPACE/artifacts
+cat << EOF > $WORKSPACE/artifacts/junit-dummy.xml
+<testsuite tests="1">
+    <testcase classname="dummy" name="dummytest"/>
+</testsuite>
+EOF
```
gwenneg, line 31: Same question.

.rhcicd/build_deploy.sh
josejulio: Yes, as of right now this is required. It might be removed/changed in the future

gburges: yes, it is needed, because if the xml doesn't exist, the jenkins job fails

.rhcicd/pr_check.sh
gburges: Same, it's needed for the jenkins job to succeed

.rhcicd/build_deploy.sh
gwenneg: Ok I thought this was only needed when JUnit was involved in the build.

BlakeHolifield (approved):

PR #246, opened 2022-03-09
Adding frontend CRD

gburges:


BlakeHolifield:
Looks good! It's on hold until the container image is built. We need the webhooks.

.rhcicd/frontend.yaml
```diff
@@ -0,0 +1,40 @@
...
+          - v1
+      frontend:
+        paths:
+          - /apps/notifications
+      image: ${IMAGE}:${IMAGE_TAG}
+      navItems:
+        - appId: "notifications"
+          title: "notifications"
+          href: "/settings/notifications"
+          product: "Red Hat Insights"
```
josejulio, line 25: This app also provides `/settings/integrations` 

Is this for the navbar? There are also submenus.

See https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/chrome/settings-navigation.json#L47-L72 for the integrations and submenus

BlakeHolifield: We'll add `integrations` as well. I thought it was a separate app. As for the submenus, that will get updated in another central PR since the nav gets shared by other resources.

josejulio:
@BlakeHolifield not directly related to this PR, but i see the following in the test:
```
08:48:42 You need to run "nvm install 16" to install it before using it.
```

gburges:
damn timeouts

/retest

.rhcicd/ci.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
```
BlakeHolifield, line 2: Can we try adding `yarn install --immutable` here? I think we just haven't installed `npm-run-all` yet.

BlakeHolifield (approved):

josejulio (approved):
lgtm!

PR #249, opened 2022-03-11
Mention local env earlier

pilhuhn:


README.md
```diff
@@ -7,6 +7,8 @@ Notifications frontend for Red Hat Insights
 
 ## Build app
 
+0. If needed run `yarn install` to install dependency packages
```
josejulio, line 10: `yarn` is an alias for `yarn install`

README.md
```diff
@@ -7,6 +7,8 @@ Notifications frontend for Red Hat Insights
...
+
 1. ```yarn```
```
josejulio, line 12:
```suggestion
1. If needed run `yarn` or `yarn install` to install dependency packages
```

josejulio (changes requested):

README.md
pilhuhn: I mention it like this, because I did a yarn and then yarn start as described. Yarn start failed. Then only as I gave yarn install it installed the dependencies and also env-cmd, which was blocking the yarn start startup.

josejulio: yarn should do an install

```bash
$ yarn
yarn install v1.22.17
[1/4] Resolving packages...
success Already up-to-date.
```

josejulio: Anyway is not a big deal to have that step in there

josejulio (approved):

PR #253, opened 2022-03-17
NOTIF-397 Disable repeated Integration actions in Behavior Groups

Yasin-Darar:
https://user-images.githubusercontent.com/80718590/159270821-7ff72ee6-30de-4224-99fc-67303780f50e.mov

### DIFF
--- a/src/components/Notifications/Form/IntegrationRecipientTypeahead.tsx
+++ b/src/components/Notifications/Form/IntegrationRecipientTypeahead.tsx
@@ -59,12 +59,12 @@ export const IntegrationRecipientTypeahead: React.FunctionComponent<IntegrationR
         }
     }, [ getIntegrations, props.integrationType, state.loadingFilter, state.lastSearch, dispatchers ]);
 
-    const getExistingIntegrations = React.useCallback(() => {
+    const existingIntegrations = React.useMemo(() => {
         const integrationActions = ((values.actions ?? []) as readonly ActionIntegration[]).map(action => action?.integration.id);
         return new Set<string>(integrationActions);
     }, [ values ]);
 
-    const options = useRecipientOptionMemo(state, getExistingIntegrations());
+    const options = useRecipientOptionMemo(state, existingIntegrations);
 
     const onFilter = React.useCallback((e: React.ChangeEvent<HTMLInputElement> | null) => {
         // Ignore filter calls with null event
--- a/src/components/Notifications/Form/IntegrationRecipientTypeahead.tsx
+++ b/src/components/Notifications/Form/IntegrationRecipientTypeahead.tsx
@@ -1,11 +1,12 @@
 import { Select, SelectOptionObject, SelectVariant } from '@patternfly/react-core';
 import { OuiaComponentProps } from '@redhat-cloud-services/insights-common-typescript';
+import { useFormikContext } from 'formik';
 import * as React from 'react';
 import { usePrevious } from 'react-use';
 
 import Config from '../../../config/Config';
 import { UserIntegrationType } from '../../../types/Integration';
-import { IntegrationRef } from '../../../types/Notification';
+import { ActionIntegration, BehaviorGroup, IntegrationRef } from '../../../types/Notification';
 import { IntegrationRecipient } from '../../../types/Recipient';
 import { getOuiaProps } from '../../../utils/getOuiaProps';
 import { useRecipientContext } from '../RecipientContext';
@@ -25,7 +26,9 @@ export interface IntegrationRecipientTypeaheadProps extends OuiaComponentProps {
 export const IntegrationRecipientTypeahead: React.FunctionComponent<IntegrationRecipientTypeaheadProps> = (props) => {
     const [ isOpen, setOpen ] = React.useState(false);
     const prevOpen = usePrevious(isOpen);
+
     const { getIntegrations } = useRecipientContext();
+    const { values } = useFormikContext<Partial<BehaviorGroup>>();
 
     const [ state, dispatchers ] = useTypeaheadReducer<IntegrationRecipient>();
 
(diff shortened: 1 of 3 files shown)


src/components/Notifications/Form/IntegrationRecipientTypeahead.tsx
```diff
@@ -53,7 +59,12 @@ export const IntegrationRecipientTypeahead: React.FunctionComponent<IntegrationR
         }
     }, [ getIntegrations, props.integrationType, state.loadingFilter, state.lastSearch, dispatchers ]);
 
-    const options = useRecipientOptionMemo(state);
+    const getExistingIntegrations = React.useCallback(() => {
```
josejulio: This could be an `useMemo`. That way we wouldn't need to call it on every render and the value would be stable.

src/components/Notifications/Form/__tests__/IntegrationRecipientTypeahead.test.tsx
```diff
@@ -150,4 +150,23 @@ describe('src/components/Notifications/Form/IntegrationRecipientTypeAhead', () =
 
         expect(onSelected).toHaveBeenCalled();
     });
+
+    it('Integration recipients that have been previously used in the form are disabled', async () => {
+        // render(<IntegrationRecipientTypeahead
```
josejulio: You could probably wrap this with `<Formik>`, and use an initial state (a behavior group that has the actions that we want to test).

See the [behavior group save modal](https://github.com/RedHatInsights/notifications-frontend/blob/9d3010e92951234db43e98892a935cc51f47a980/src/components/Notifications/BehaviorGroup/BehaviorGroupSaveModal.tsx#L60).

josejulio (commented):
Looking good!

josejulio:
I know that is not finished, but when you are done, could you add a gif showing the change? It helps others to review the visual changes.

src/components/Notifications/Form/IntegrationRecipientTypeahead.tsx
Yasin-Darar: My understanding was that useCallback memoizes the return value while useMemo memoizes the function itself, would useCallback not be the correct one in this instance?

josejulio: [useCallback](https://reactjs.org/docs/hooks-reference.html#usecallback) will memoize the function itself - This is useful when we want to pass the function as an argument to other component or when we want to use the the function as a dependency.

[useMemo](https://reactjs.org/docs/hooks-reference.html#usememo) will memoize the result of the function itself.

As found in the docs:
> `useCallback(fn, deps) is equivalent to useMemo(() => fn, deps).`

Yasin-Darar: Ok, it seems that I had them backwards.

src/components/Notifications/Form/__tests__/IntegrationRecipientTypeahead.test.tsx
```diff
@@ -152,21 +165,22 @@ describe('src/components/Notifications/Form/IntegrationRecipientTypeAhead', () =
...
+            onSelected={ fn() }
+        />, {
+            wrapper: getConfiguredWrapper(undefined, formikValues)
+        });
+
+        await waitForAsyncEvents();
+        userEvent.click(screen.getByRole('button', { name: /Options menu/i }));
+
+        await waitForAsyncEvents();
+        expect(screen.getAllByRole('option')[0].className).toEqual('pf-c-select__menu-item pf-m-disabled');
```
josejulio: Could we check if the following attribute is present? `aria-disabled` (and set to true) instead of the className?

Yasin-Darar: Unfortunately, I could not find a cleaner way to check if it was disabled. It does not have an aria-disabled attribute and when I used toBeDisabled, it also was complaining.

josejulio: Yes, they usually don't really disable it and they add a `aria-disabled`.
Maybe this is a bug in patternfly. Could you create an issue here? 
https://github.com/patternfly/patternfly-react

josejulio (approved):

josejulio:
@jeromemarc @mmenestr 
This PR prevents from selecting the same integration multiple times in the same behavior group by disabling it

@Yasin-Darar could you add a screenshot showing the change?

Yasin-Darar:
> @jeromemarc @mmenestr This PR prevents from selecting the same integration multiple times in the same behavior group by disabling it
> 
> @Yasin-Darar could you add a screenshot showing the change?

Do you see the video I embedded at the top?

josejulio:
> Do you see the video I embedded at the top?

I do, thanks!

josejulio:
I wonder if we should add some sort of text to let the user know why they can't select the integration.

Yasin-Darar:
> I wonder if we should add some sort of text to let the user know why they can't select the integration.

What do you think about a tooltip that says something like "This integration has already been added"?

josejulio:
> What do you think about a tooltip that says something like "This integration has already been added"?

Seems good to me

Yasin-Darar:
@josejulio I just pushed updated changes to add the text indication when integrations are disabled (video should be updated).

josejulio (approved):
lgtm

josejulio:
We can merge when the tests are green

theute:
@Yasin-Darar congrats on your first merged PR here :-)

mmenestr:
Hey, sorry I didn't get a chance to review this before the merge but it looks great! Thank you for the tag and for adding the GIF!! If you ever need a review and I take too long to give one, feel free to ping me directly with the PR link :) Or ping me if you ever have any UX questions about anything @Yasin-Darar

PR #255, opened 2022-03-22
NOTIF-326: Searching Empty State for Event Logs

Yasin-Darar:
https://user-images.githubusercontent.com/80718590/159554342-1508863e-06a3-4cb1-82ee-4832125ad133.mov

### DIFF
--- a/src/components/EmptyStateSearch.tsx
+++ b/src/components/EmptyStateSearch.tsx
@@ -1,0 +1,27 @@
+import { EmptyState, EmptyStateBody, EmptyStateIcon, EmptyStateVariant, Title } from '@patternfly/react-core/dist/js/components';
+import { SearchIcon } from '@patternfly/react-icons';
+import React from 'react';
+import { style } from 'typestyle';
+
+interface ParentProps {
+    className?: string,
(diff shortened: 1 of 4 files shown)


src/components/EmptyStateSearch.tsx
```diff
@@ -0,0 +1,27 @@
+import { EmptyState, EmptyStateBody, EmptyStateIcon, EmptyStateVariant, Title } from '@patternfly/react-core/dist/js/components';
```
josejulio: We should import from the base of the library. autocomplete messes it up sometimes.

```suggestion
import { EmptyState, EmptyStateBody, EmptyStateIcon, EmptyStateVariant, Title } from '@patternfly/react-core';
```

josejulio (changes requested):
Looks good, one detail I found.

src/components/EmptyStateSearch.tsx
Yasin-Darar: Ok, that has been changed, resolving thread.

josejulio (approved):
lgtm!

PR #257, opened 2022-03-24
Fix: calling undefined/wrong type object

josejulio:
This was happening when opening a behavior group that had both email and a integration. There were other scenarios that triggered this error - creating a new behavior group and adding emails and integrations also made this fail.


Yasin-Darar (approved):
lgtm

PR #259, opened 2022-03-24
NOTIF-323: Change Filter By Application to Tree Design

Yasin-Darar:
https://user-images.githubusercontent.com/80718590/161845349-d235c668-0d98-4f94-bf6a-fe1c4fbf15fc.mov


tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es2019",
-    "isolatedModules": true,
+    "isolatedModules": false,
```
josejulio: Why did you need to change the `isolatedModules` to `false` ?

src/hooks/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,92 @@
...
+            name: item.displayName,
+            checkProps: { checked: false }
+        }))
+
+    }))
+
+    console.log("render")
+    return (
+        <Dropdown
+            onSelect={() => setIsFocused(!isFocused)}
```
josejulio: I don't think you want this here. 
The `onSelect` is used to know when you clicked something from the content.
Since you are going to use a tree-view; that component should handle their own selection/clicks.

```suggestion
```

src/hooks/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,92 @@
...
+            toggle={
+                <DropdownToggle toggleIndicator={AngleDownIcon}>{props.placeholder}</DropdownToggle>
```
josejulio: This is the toggle (the visible part of the Dropdown when is closed).
Here is where you want to control the behavior of the opening/closing.

```suggestion
                <DropdownToggle onToggle={ () => setIsFocused(prev => !prev) } toggleIndicator={AngleDownIcon}>{props.placeholder}</DropdownToggle>
```

tsconfig.json
Yasin-Darar: I use JSX in a hook file I am creating and it didn't like that it was not an isolated module.. I disabled the option so I could continue to implement/test the feature but I will change it back when I move this MR out of Draft.

src/hooks/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: Thank you, that fixes my issue... pretty careless mistake of me. I totally forgot about the component used specifically for toggling the dropdown.

tsconfig.json
josejulio: FWIW exporting something should make it an "isolated module"

src/hooks/usePrimaryToolbarFilterConfigWrapper.tsx
josejulio: Nothing to worry, It happens all the time - an extra pair of eyes is always helpful.

tsconfig.json
Yasin-Darar: I tried doing `export {}` as a temporary workaround but it didn't fix it... maybe I needed to restart my development server for it to take into effect.

src/components/Notifications/EventLog/EventLogFilter.ts
```diff
@@ -1,9 +1,9 @@
 import { ClearFilters, Filters, SetFilters } from '@redhat-cloud-services/insights-common-typescript';
 
 export enum EventLogFilterColumn {
-    EVENT = 'event',
     APPLICATION = 'application',
-    BUNDLE = 'bundle'
+    EVENT = 'event',
+    BUNDLE = 'bundle',
```
josejulio, line 6: We can drop this one, won't be used.

src/components/Notifications/EventLog/EventLogToolbar.tsx
```diff
@@ -39,44 +38,19 @@ interface EventLogToolbarProps extends OuiaComponentProps {
 }
 
 export const EventLogToolbar: React.FunctionComponent<EventLogToolbarProps> = (props) => {
-
-    const filterMetadata = React.useMemo<ColumnsMetada<typeof EventLogFilterColumn>>(() => {
-        const bundleOptions = props.bundleOptions;
-        const applicationOptions = props.applicationOptions;
-        return {
+    const filterMetadata = React.useMemo<ColumnsMetada<any>>(() => {
```
josejulio: Could we take off that `any` ? 
Maybe something like: Partial<ColumnsMetada<typeof EventLogFilterColumn>>

src/pages/Notifications/EventLog/useFilterBuilder.ts
```diff
@@ -23,8 +22,21 @@ export const useFilterBuilder = (
...
         if (filters?.application) {
-            const selectedAppNames = filters?.application;
-            const selectedApps = applications.filter(a => selectedAppNames.includes(a.name)).map(a => a.id);
+            const selectedBundleNames = filters.bundle as string[];
+            const selectedAppNames = filters.application as string[];
+
+            const selectedApps: string[] = [];
+            selectedBundleNames.forEach(selectedBundle => {
+                const bundle = bundles.find(bundle => bundle.name === selectedBundle);
+                if (!!bundle) {
```
josejulio: nitpick: I think you can directly do `if(bundle)` - a boolean type is not required

src/pages/Notifications/EventLog/useFilterBuilder.ts
```diff
@@ -23,8 +22,21 @@ export const useFilterBuilder = (
...
+                    selectedAppNames.forEach(selectedApp => {
+                        const application = bundle.children?.find(application => application.name === selectedApp);
+                        if (!!application) {
```
josejulio: nitpick: same as above

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+                        const activeChips = activeFilter.chips.map(chip => chip.value);
+
+                        treeNode.children?.forEach(childNode => {
+                            childNode.checkProps.checked = activeChips.includes(childNode.id);
+                        });
+
+                        if (allChildrenChecked(treeNode)) {
+                            treeNode.checkProps.checked = true;
+                        }
+                        else if (childChecked(treeNode)) {
```
josejulio: nitpick: I tend to group my if-else blocks in this way:
```
if (foo) {
} else if (bar) {
} else {
}
```

Nothing critical, but we could agree on something to have it consistent  across the code  (and maybe update the linter to enforce it)

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+                        }
+                        else {
+                            treeNode.checkProps.checked = false;
+                        }
+                    });
+                }
+            }));
+        }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [ groups, filters ]);
```
josejulio: Is this complaining only about the `initialize`? What would happen if it's added there? 
if something is breaking, I would update the `initialize` to not depend on anything and put the `eslint-disable` there. This function is more complex and requires more attention.

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+export const EventLogTreeFilter: React.FunctionComponent<EventLogTreeFilterProps> = (props) => {
+    const { groups, placeholder, filters, updateFilters } = props;
+
+    const initialize = React.useMemo(() => initTreeNodeById(groups, filters), [ groups, filters ]);
+
+    const [ treeNodeById, setTreeNodeById ] = React.useState<TreeNodeDict>(initialize);
+    const [ isToggled, setIsToggled ] = React.useState(false);
+
+    const treeDataArray = React.useMemo(() => !!treeNodeById ? Object.values(treeNodeById) : [], [ treeNodeById ]);
```
josejulio: nitpick
```suggestion
    const treeDataArray = React.useMemo(() => treeNodeById ? Object.values(treeNodeById) : [], [ treeNodeById ]);
```

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+                }
+            }));
+        }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [ groups, filters ]);
+
+    const flattenTree = React.useMemo(() => {
+        const flatTreeDataArray = produce(treeDataArray, (prev) => {
+            prev.forEach(treeNode => {
+                if (!!treeNode.children) {
```
josejulio: nitpick:
```suggestion
                if (treeNode.children) {
```

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+        });
+
+        return flatTreeDataArray;
+    }, [ treeDataArray ]);
+
+    React.useEffect(() => {
+        const groupNames = groups.map(group => group.name);
+        const activeParentFilters = flattenTree.filter(treeNode => {
+            const isActive = treeNode.checkProps.checked;
+            return (!!isActive || isActive === null) && groupNames.includes(treeNode.id);
```
josejulio:
```suggestion
            return (isActive || isActive === null) && groupNames.includes(treeNode.id);
```

Also, why `isActive` is null?

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+        });
+
+        updateFilters(produce((prev) => {
+            if (prev.length === 0 && activeParentFilters.length === 0) {
+                return prev;
+            }
+
+            else if (prev.length !== 0 && activeParentFilters.length === 0) {
```
josejulio:
```suggestion
            } else if (prev.length !== 0 && activeParentFilters.length === 0) {
```
or at least lest keep them closer

src/components/Notifications/EventLog/EventLogToolbar.tsx
```diff
@@ -39,44 +38,19 @@ interface EventLogToolbarProps extends OuiaComponentProps {
...
+        const metaData: ColumnsMetada<any> = {
```
Yasin-Darar: Instead of putting into variable, just return directly from the memo since it is just constant

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+    const [ treeNodeById, setTreeNodeById ] = React.useState<TreeNodeDict>(initialize);
+    const [ isToggled, setIsToggled ] = React.useState(false);
+
+    const treeDataArray = React.useMemo(() => !!treeNodeById ? Object.values(treeNodeById) : [], [ treeNodeById ]);
+
+    React.useEffect(() => {
+        if (groups.length !== 0) {
+            setTreeNodeById(produce((prev) => {
+                // In case Bundles request is pending during Tree View render
```
Yasin-Darar: misleading comment, remove

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+        updateFilters(produce((prev) => {
+            if (prev.length === 0 && activeParentFilters.length === 0) {
+                return prev;
+            }
+
+            else if (prev.length !== 0 && activeParentFilters.length === 0) {
+                return [];
+            }
+
```
Yasin-Darar, line 140: add else conditional for readability

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+                            return true;
+                        }
+
+                        return false;
+                    });
+                });
+
+                idxToRemove.forEach((indexRemove, idx) => {
+                    prev.splice(indexRemove - idx, 1);
```
Yasin-Darar: Mutate inside forEach (tested in sandbox, forEach is smart enough to index the pre computed length of array and idx param also changes accordingly)

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+            const newCustomFilters = activeParentFilters.map(parentFilter => ({
+                bundleId: parentFilter.id,
+                category: parentFilter.name as string,
+                chips: parentFilter.children?.filter(childNode => !!childNode.checkProps.checked).map(childFilter => ({
+                    name: childFilter.name as string,
+                    value: childFilter.id,
+                    isRead: true
+                })) ||
```
josejulio: Could be replaced with the [Nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)
In this case it's more or less the same, but might help when reading the code.
```suggestion
                })) ??
```

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+                    {
+                        name: parentFilter.name,
+                        value: parentFilter.id,
+                        isRead: true
+                    }
+                ]
+            } as EventLogCustomFilter));
+
+            const areEqual = newCustomFilters.every((entry, idx) => {
+                if (!!prev[idx]) {
```
josejulio:
```suggestion
                if (prev[idx]) {
```

src/components/Notifications/EventLog/EventLogFilter.ts
Yasin-Darar: I was not sure if we still wanted to keep it for the URL filtering... right now, I have it so that if all the applications are selected for a specific bundle, the bundle shows up in the URL instead of the corresponding applications. I think it is easier to understand IMO. 

However, if not all the applications are listed then I list the bundle and it's applications. This approach does have a drawback, which is since different bundles can have same application names, if you had two bundles selected in the URL and two applications that were present in both bundles, it might filter out the wrong ones depending on which one comes first in the bundles network request.

src/pages/Notifications/EventLog/useFilterBuilder.ts
Yasin-Darar: Yeah, I tend to use the double bang operator for easier readability but I can remove it since conditional statements already check for truthy values

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
Yasin-Darar: I am definitely in the minority when it comes to grouping my conditional blocks... I think having something consistent across the codebase is the most important part.. unfortunately, i have too much muscle memory to do it that way to begin with. Will change to suggested structure and second the idea of having the linter enforce it, because I know I'll forget again.

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
Yasin-Darar: For the TreeView patternfly component, null signifies a "half checked state" which means if any of the children are checked it displays a 'minus' sign to signify that it is being applied but not all of the children filters. 

Quick image of what I'm talking about.

image

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
Yasin-Darar: I will change to aforementioned style as stated in my reply above.

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
Yasin-Darar: Yeah, in this case, they would be doing the same thing... but null coalescing operator is more properly scoped since I only care about undefined cases in this snippet instead of all falsy values. Will change

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
Yasin-Darar: This function has undergone a lot of changes over the course of this ticket... I think I added that disable for a previous commit where it depended on something that was causing a circular dependency... but it seems that initialize is the only thing that is missing and initialize depends on the same things as this effect, so in theory it should be fine if I were to add the dependency. 

Will investigate further to ensure no regression.

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+            });
+
+            return areEqual ? prev : newCustomFilters;
+        }));
+    }, [ groups, flattenTree, updateFilters ]);
+
+    const onCheck = (event: ChangeEvent<Element>, treeNode: TreeNodeItem, parentNode: TreeNodeItem) => {
+        const checked = (event.target as HTMLInputElement).checked;
+        setTreeNodeById(produce((prev) => {
+            if (!!parentNode) {
```
josejulio:
```suggestion
            if (parentNode) {
```

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+                    return false;
+                });
+
+                if (allChildrenChecked(prev[parentNode.id])) {
+                    prev[parentNode.id].checkProps.checked = true;
+                }
+                else if (childChecked(prev[parentNode.id])) {
+                    prev[parentNode.id].checkProps.checked = null;
+                }
+                else {
```
josejulio:
```suggestion
                } else if (childChecked(prev[parentNode.id])) {
                    prev[parentNode.id].checkProps.checked = null;
                } else {
```

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+                    prev[parentNode.id].checkProps.checked = checked;
+                }
+            }
+            else {
```
josejulio:
```suggestion
            } else {
```

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,214 @@
...
+        <Dropdown
+            toggle={ <DropdownToggle
+                onToggle={ () => setIsToggled(!isToggled) }
+                toggleIndicator={ AngleDownIcon }
+            >
+                {placeholder}
+            </DropdownToggle> }
+            isOpen={ isToggled }
+        >
+            <TreeView data={ treeDataArray } hasChecks={ true } onCheck={ onCheck as any } />
```
josejulio: What's the complain of the `onCheck`  (i.e. why the cast to `any`) ?

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,180 @@
...
+    );
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    const defaultDelete = React.useMemo(() => toolbarConfig.activeFiltersConfig.onDelete, []);
+    toolbarConfig.activeFiltersConfig.onDelete = React.useCallback((
+        _event: any,
+        rawFilterConfigs: any[]
+    ) => {
+        try {
+            defaultDelete(_event, rawFilterConfigs);
```
josejulio: Instead of this approach, could we try: 
. Detect if the `rawFilterConfigs` has any detail that is important for us.
  . If has any, process and strip that from the `rawFilterConfigs` (create a new object)
. process the `defaultDelete`

My worry is that the catch would swallow other errors that would go unnoticed

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,180 @@
...
+    const toolbarConfig = usePrimaryToolbarFilterConfig(
+        EventLogFilterColumn,
+        filters,
+        setFilters,
+        clearFilter,
+        metaData
+    );
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    const defaultDelete = React.useMemo(() => toolbarConfig.activeFiltersConfig.onDelete, []);
```
josejulio: Should probably add `toolbarConfig.activeFiltersConfig.onDelete` as a dependency.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,180 @@
...
+            };
+        }) || [];
+
+        return eventLogCustomFilters;
+    }, []);
+
+    React.useEffect(() => {
+        setCustomFilters(mapToEventLogCustomFilter(filters, bundles.length !== 0 ? bundles : undefined));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [ bundles, mapToEventLogCustomFilter ]);
```
josejulio, line 151: missing filters

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+    toolbarConfig.filterConfig.items[1] = useMemo(() => {
```
josejulio: It's always on the position `1` ?

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,180 @@
...
+            }
+        } as any;
+    }, [ bundles, customFilters ]);
+
+    toolbarConfig.activeFiltersConfig.filters = React.useMemo(() => {
+        const activeFilters = toolbarConfig.activeFiltersConfig.filters;
+        const nonCustomFilters = activeFilters.filter(activeFilter => !!activeFilter && !(activeFilter as EventLogCustomFilter).bundleId);
+        return nonCustomFilters.concat(customFilters);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [ filters, customFilters ]);
```
josejulio: missing `toolbarConfig.activeFiltersConfig.filters` .

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,180 @@
...
+                    if (!activeApplications.includes(chipValue)) {
+                        activeApplications.push(chipValue);
+                    }
+                });
+            }
+        });
+
+        return { activeBundles, activeApplications };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [ bundles, customFilters ]);
```
josejulio: Check missing dependencies, usually we don't need to skip this lint rule.

src/components/Notifications/EventLog/EventLogFilter.ts
josejulio: > I was not sure if we still wanted to keep it for the URL filtering... right now, I have it so that if all the applications are selected for a specific bundle, the bundle shows up in the URL instead of the corresponding applications. I think it is easier to understand IMO.

We still want to have the URL filtering; if we do require this for keeping it, then it's fine.

> However, if not all the applications are listed then I list the bundle and it's applications. This approach does have a drawback, which is since different bundles can have same application names, if you had two bundles selected in the URL and two applications that were present in both bundles, it might filter out the wrong ones depending on which one comes first in the bundles network request.

I see, maybe we want to prepend the applications with the bundle and skip the bundle altogether ? I'm not really sure because the url could probably grow a lot. Lets try solving that in a separate PR

josejulio:
Lots of the review comments are style; perhaps our linter should take care of most of all. I suppose we need to turn on something to ensure that.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: I purposefully left out filters because it causes a circular dependency

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: It can technically be at any position, I just have it always at 1

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
Yasin-Darar: I use my own typing for the params of the onCheck, so it is complaining because the function param types are different

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: I only care about what onDelete on mount... I actually do not want this to be called again because I reassign `toolbarConfig.activeFiltersConfig.onDelete` on the next line... so it actually ends up doing 4 recursion calls until the onDelete is pointed back to the original function.

Yasin-Darar: It's more efficient ignoring that dependency

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: Do not remember exactly why I didn't add this one, I did light testing and it seems that everything still works fine with it appended. In theory, 'activeFiltersConfig.filters' would only ever change due to 'filters' or 'customFilters' so this dependency should not have an effect on it, so it would not hurt to add it.

Yasin-Darar:
> Lots of the review comments are style; perhaps our linter should take care of most of all. I suppose we need to turn on something to ensure that.

Long term that makes the most sense, but addressing those in early PRs will lead to a lot less instances of those happening later down the line. It would definitely help ensure consistent styling and save some time in the code review portion, but I don't find the issue that alarming.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: Instead of doing it this way, which is using a memo to memoize the current filtered bundles and applications to keep the URL params up to date and having to omit dependencies... I will use `produce` function inside the `setFilters.bundle` and `setFilters.application`.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: That's a good catch, however, I've looked through the original onDelete function and there are no errors that are thrown outside of `unexpected column/filter label` which are derived on the metadata. Since the custom filter is not in the metadata it has no purpose calling that function.

But this point does bring up the case where the existing 'non custom' filters (that do exist in the metadata) cause an error... the current implementation will assume that it was a custom filter causing the error, which is wrong. I will change to accommodate that, but I don't see a reason for the custom filter deletes to call the original one.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: I may be able to add the dependency now that I restructures the `setFilters` code, will test and evaluate.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: Actually, now I understand what you are referring to, if you were to clear filters with a mix of custom and non custom, you would need to go through both delete functionality

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: After testing, I have verified that the current implementation prevents the `filters` dependency from being added.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: I have a ticket that adds some filters for EventLog ([NOTIF-324](https://issues.redhat.com/browse/NOTIF-324) I think), I will make this dynamic under that PR... resolving for now.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: I changed this design completely because of this comment, so this is no longer relevant, resolving....

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: Implemented more robust solution following the pseudo code in the parent comment

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,207 @@
...
+                            isRead: true
+                        }
+                    ]
+                } as EventLogCustomFilter));
+
+                const prevBundles = prev.map(prevFilter => prevFilter.bundleId);
+                const currBundles = newCustomFilters.map(customFilter => customFilter.bundleId);
+
+                const areBundlesEqual = areEqual(prevBundles, currBundles);
+                const areApplicationsEqual = newCustomFilters.every((entry, idx) => {
```
Yasin-Darar: add `areBundlesEqual` to this line to prevent unnecessary iterations if we know the filters are not equal

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
josejulio: Could we try to adapt to their `onCheck`  type?
They could update the signature and change the order of the params or what not in a major update - thus breaking without us noticing. If you need something custom, you can wrap it up.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
josejulio: Unresolving to take a second look here.

Yasin-Darar: When adding the dependency, it causes 8 invocations of the useMemo during the initial rendering of the Event Log Page. Then it causes another several recursive calls to occur when deleting since the memory address of the delete function gets pointed now at the wrapper function (essentially pointing at itself). I think it breaks out of the infinite loop at some point due to another invocation of the useMemo that resets it. That is my speculation at least, not 100% sure why it does not end in a stack overflow from infinite recursive calls. 

Either way, it seems that it would still work if you put in the dependency but it would make it worse in terms of efficiency.

josejulio: I just wanna double check that the stale dependency won't be a source of trouble

Yasin-Darar: Have you gotten a chance to check this out?

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+export const EventLogTreeFilter: React.FunctionComponent<EventLogTreeFilterProps> = (props) => {
+    const { groups, placeholder, filters, updateFilters } = props;
+
+    const initialize = React.useMemo(() => initTreeNodeById(groups, filters), [ groups, filters ]);
+
+    const [ treeNodeById, setTreeNodeById ] = React.useState<TreeNodeDict>(initialize);
+    const [ isToggled, setIsToggled ] = React.useState(false);
+
+    const treeDataArray = React.useMemo(() => treeNodeById ? Object.values(treeNodeById) : [], [ treeNodeById ]);
```
josejulio: Isn't `treeNodeById` always truthy? What does the typecheck says?

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
```diff
@@ -0,0 +1,211 @@
...
+        });
+
+        return flatTreeDataArray;
+    }, [ treeDataArray ]);
+
+    React.useEffect(() => {
+        const groupNames = groups.map(group => group.name);
+        const activeParentFilters = flattenTree.filter(treeNode => {
+            const isActive = treeNode.checkProps.checked;
+            return (isActive || isActive === null) && groupNames.includes(treeNode.id);
```
josejulio: What's the format of id? Is it like `the-group-id.the-child-id` ? If so maybe we can change this to something like:
```suggestion
            return (isActive || isActive === null) && treeNode.id.startsWith(`${groupNames}.`);
```

to avoid picking up false positives from the include in case the app name is similar to a bundle.

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
josejulio: Something to consider. I see that this `treeDataArray` is used to only generate `flattenTree`  which is then used to only generate a `activeParentFilters`  which is then used to generate a `newCustomFilters`. 

I guess we could simplify all that into a single step, and the `newCustomFilters` could be stable enough to do a ref check (i.e. old == new)

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
josejulio: I guess the issue then is why `toolbarConfig.activeFiltersConfig.onDelete` is so unstable - If it's not using the latest, then the `defaultDelete` is not really the latest and we might be using old code in the next function.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,230 @@
...
+        EventLogFilterColumn,
+        filters,
+        setFilters,
+        clearFilter,
+        metaData
+    );
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    const defaultDelete = React.useMemo(() => toolbarConfig.activeFiltersConfig.onDelete, []);
+    toolbarConfig.activeFiltersConfig.onDelete = React.useCallback((
```
josejulio: I didn't notice this before, but we should create a new object instead of modifying what the previous hook gave us.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: but why would the code change during the execution of the filter? If it gets set on mount each time, it's not like the code for the delete would change during that time.

Yasin-Darar: Does the instability of the onDelete have to do with reassigning it on the next line under?
Screen Shot 2022-04-08 at 11 13 02 AM

josejulio: Ah - bingo, it surelly does.

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
Yasin-Darar: groupId.childId format is only used for the URL query param... the treeNodes carry just the id of whatever thing it is representing, so app would be appId... bundle would be bundleId

Yasin-Darar: So it is possible to have an application with the same name as a bundle?

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
Yasin-Darar: I think I changed treeNodeById to not be initialized to undefined at the start during some refactors from other comments and forgot to update the memo. 

I will look into having all the steps for generating a new filter list be condensed.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: would wrapping everything in `produce(toolbarConfig, () =>{curriedFunc})` be the best solution for that or simply using the `...` spread operator?

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
josejulio: We don't have that now, but not imposible to have some sort of clash in the future think of a group called: "application" and an app called "application-services". We don't enforce what an app or a bundle can be called so we can't make the assumption that it won't happen.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
josejulio: a produce would be good. We can put all our updates in variables and modify at the end.

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
Yasin-Darar: groupNames is an array, it would only return true if any instance in the string array matches exactly with the instance of the treeNode id.... so they would have to be an exact match in terms of names. I think there are other areas that would need some major refactoring if an assumption was that app names and bundle names could be the same.

josejulio: ah - nevermind my comment then :) for a moment thought that `groupNames` was a string.

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
Yasin-Darar: I use `treeDataArray` also as the data to pass into the props for `TreeView` component... I think having that consistently memoized is beneficial.

I removed the need for flattenTree, so now instead of the effect depending on `flattenTree` it depends on `treeDataArray` which simplifies the logic a little bit. Is your comment referring to a method to remove the data altogether because I'm not too sure that is possible with the current design.

josejulio: I'm not asking to remove, just to condense into a single memo when it makes sense. 
I think it would be worth to take this out and put in a memo (again, it might be condensed with a previous step)
https://github.com/RedHatInsights/notifications-frontend/pull/259/files#diff-b347d923933f0f81d5d65b68fb37caf1d7070f3d58871e06d86144ea3823eec1R126

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: Will change to use dependency since the stability of `onDelete` will be preserved from the refactor that is mentioned in other comment

src/components/Notifications/EventLog/EventLogTreeFilter.tsx
Yasin-Darar: That is a good suggestion, I could extrapolate that into a memo since it only depends on treeDataArray

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,234 @@
...
+    }, [ bundles, customFilters ]);
+
+    const activeFiltersConfig = React.useMemo(() => {
+        const activeFilters = toolbarConfig.activeFiltersConfig.filters as EventLogCustomFilter[];
+        const nonCustomFilters = activeFilters.filter(activeFilter => activeFilter && !activeFilter.bundleId);
+        return nonCustomFilters.concat(customFilters);
+    }, [ customFilters, toolbarConfig.activeFiltersConfig.filters ]);
+
+    // Update URL Query Params for Bundles
+    setFilters.bundle(produce(prev => {
```
josejulio: Is this executing on every call? I wonder if this should be inside a useEffect that depends on `setFilters.bundle` and `bundles`

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,234 @@
...
+            if (addToQueryParam) {
+                currBundleFilters.push(bundle.name);
+            }
+        });
+
+        return areEqual(prev as string[], currBundleFilters, true) ? prev : currBundleFilters;
+    }));
+
+    // Update URL Query Params for Bundles
+    setFilters.application(produce(prev => {
```
josejulio: Is this executing on every call? I wonder if this should be inside a useEffect that depends on `setFilters.application` and `bundles`

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,234 @@
...
+    const activeFiltersConfig = React.useMemo(() => {
+        const activeFilters = toolbarConfig.activeFiltersConfig.filters as EventLogCustomFilter[];
+        const nonCustomFilters = activeFilters.filter(activeFilter => activeFilter && !activeFilter.bundleId);
+        return nonCustomFilters.concat(customFilters);
+    }, [ customFilters, toolbarConfig.activeFiltersConfig.filters ]);
+
+    // Update URL Query Params for Bundles
+    setFilters.bundle(produce(prev => {
+        if (bundles.length === 0) {
+            return filters.bundle;
```
josejulio: Is this meant to be a no-op?  You can return `undefined` (from an immer produce) to mark "nothing happened"

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -0,0 +1,234 @@
...
+            }
+        });
+
+        return areEqual(prev as string[], currBundleFilters, true) ? prev : currBundleFilters;
+    }));
+
+    // Update URL Query Params for Bundles
+    setFilters.application(produce(prev => {
+        if (bundles.length === 0) {
+            return filters.application;
```
josejulio: Is this meant to be a no-op?  You can return `undefined` (from an immer produce) to mark "nothing happened"

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: Since the memory address is not changing each time, it doesnt cause any new invocations of hooks unless something has been changed. 

I guess it technically does not need to be called every time. It really depends on the performance difference of not calling it vs calling it with the same object as before (which I presume is negligent).

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: You are correct that is supposed to be a case where nothing changes... will change to `return` to better symbolize that, even though I think technically as it is now, it does the same thing functionally.

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
Yasin-Darar: After some more thought, I realized that each time it rerenders it needs to go through all the code in the `produce` func in order to determine if the `draft` obj is different from the `original` which would be where the inefficiency comes from, which makes sense to wrap it in a `useEffect`. Will change and test.

josejulio (approved):
lgtm

PR #261, opened 2022-03-28
NOTIF-347 Preparations for groups as recipients (email)

josejulio:
This prepares/refactor some of the code to make it easier to review future PRs with the recipient as admin.

- Fixes some immer usage (using original object instead of proxied to prevent building proxies)
- Extracts mapper of the options hook to allow each class to customize without affecting the others
- Prepares for new email recipient (group) 
  - Adds a new email recipient base
  - New type (not being used yet)

### DIFF
--- a/src/types/Recipient.ts
+++ b/src/types/Recipient.ts
@@ -32,11 +32,26 @@ export class IntegrationRecipient extends Recipient {
     }
 }
 
-export class NotificationRecipient extends Recipient {
-    readonly sendToAdmin: boolean;
+export abstract class BaseNotificationRecipient extends Recipient {
     readonly integrationId: UUID | undefined;
     readonly key: string;
     readonly description: string | undefined;
+
+    protected constructor(displayName: string, description: string, integrationId: UUID | undefined, key: string) {
+        super(displayName);
+
+        this.key = key;
+        this.description = description;
+        this.integrationId = integrationId;
+    }
+
+    public getKey() {
+        return this.key;
+    }
+}
+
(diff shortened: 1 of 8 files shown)


src/types/Recipient.ts
```diff
@@ -32,12 +32,27 @@ export class IntegrationRecipient extends Recipient {
     }
 }
 
-export class NotificationRecipient extends Recipient {
-    readonly sendToAdmin: boolean;
+export abstract class BaseNotificationRecipient extends Recipient {
```
Yasin-Darar, line 35: I'm not sure I see the reason for using an abstract class.... are you using it to prevent instantiation of BaseNotificationRecipient?

josejulio: yes - plus I don't need to implement `equals` (defined int the base). I think I could also get rid of `key` and generate it on the child classes too instead of receiving it.

Yasin-Darar: Couldn't those reasons be accomplished by using regular inheritance? Generally, I tend to see abstract for instances where you declare a class function then later define it under the inheritors (abstract functions) or for some sort of override, but here it seems to me that it's being used for normal inheritance.

josejulio: There is one abstract method that I don't want to implement right here, thus the need to make it abstract: [equals](https://github.com/RedHatInsights/notifications-frontend/blob/48ee3fc48f648cc8268870133770f1d8f20cbb82/src/types/Recipient.ts#L11)

Yasin-Darar:
LGTM

Yasin-Darar (approved):

PR #264, opened 2022-03-30
Splunk Automation Page

jpramos123:


src/Routes.tsx
```diff
@@ -19,7 +20,8 @@ interface Path {
 export const linkTo = {
     integrations: () => '/integrations',
     notifications: (bundle: string) => `/notifications/${bundle}`,
-    eventLog: (bundle?: string) => `/notifications/eventlog${bundle ? `?bundle=${bundle}` : ''}`
+    eventLog: (bundle?: string) => `/notifications/eventlog${bundle ? `?bundle=${bundle}` : ''}`,
+    splunk: () => '/integrations/splunkSetup'
```
josejulio: could you use `splunk-setup` for the url? :D

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
```diff
@@ -0,0 +1,396 @@
...
+    reduxDispatch: Dispatch;
+}
+
+export const SplunkSetupPage: React.FunctionComponent<SplunkPageProps> = props => {
+
+    const [ hecId, setHecId ] = useState('');
+    const [ splunkServerHostName, setHostName ] = useState('');
+    const [ automationLogs, setAutomationLogs ] = useState(`CLICK THE BUTTON TO START THE AUTOMATION\n`);
+
+    function handleHostInputChange(value) {
```
josejulio: Could you use arrow functions? Only asking for consistency across the code.
Also, some of these functions can be simplified:
i.e. 
```
const handleHostInputChange = setHostName; // or setHostName can be used directly
```

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    function handleHECInputChange(value) {
+        setHecId(value);
+    }
+
+    function createGroup() {
+
+        return new Promise(function (resolve, reject) {
+            axios.post(RBAC_GROUPS_API,
```
josejulio: I see that you are using await syntax elsewhere, I think you can use this syntax here as well:

```ts
const createGroup = async() => {
  try {
     const response = await axios.post(...);
     const result = response.data;
     const message = 'Group created\n';
     setAutomationLogs(`${automationLogs}${message}`);
  }  catch(error) {
    console.log('Failed to create group');
    throw error;
  }
}
```

src/pages/Integrations/SplunkSetup/SplunkSetup.scss
```diff
@@ -0,0 +1,9 @@
+
```
josejulio, line 1: Generally I try to keep the css along the code using [typestyle](https://typestyle.github.io/#/). Not a hard requirement, but would be nice if you could make use of that.

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
jpramos123: Much comfortable  syntax!

thanks for the tip

src/Routes.tsx
jpramos123: Done!

src/pages/Integrations/SplunkSetup/Constants.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const SPLUNK_GROUP_NAME = 'SPLUNK_INTEGRATION';
+export const SPLUNK_INTEGRATION_NAME = 'SPLUNK_AUTOMATION';
+export const SPLUNK_BEHAVIOR_GROUP_NAME = 'SPLUNK_AUTOMATION_GROUP';
+
+const EVENT_TYPES_PROD = [];
+const EVENT_TYPES_STAGE = [ '6d4d1e03-f1bf-427c-8c62-08a114242b5d', // policy-triggered
+    '95a6d179-5dbd-4de4-a726-5bab2a699912',  // new-recommendation
+    '5df33474-9cfd-40e9-9a86-9f8857147f89', // drift-baseline-detected
+    '75233a10-646c-4e26-a5f1-639a7df8f29f' ]; // resolved-recommendation
```
vkrizan, line 21: Is it possible to use APIs to grab those ids?

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
```diff
@@ -0,0 +1,366 @@
...
+    FormGroup,
+    Popover,
+    TextInput
+} from '@patternfly/react-core';
+import HelpIcon from '@patternfly/react-icons/dist/esm/icons/help-icon';
+import { Main, PageHeader, PageHeaderTitle, Section } from '@redhat-cloud-services/frontend-components';
+import axios from 'axios';
+import React, { useEffect, useState } from 'react';
+import { connect } from 'react-redux';
+import { Dispatch } from 'redux';
```
vkrizan, line 21: What do we need Redux for? Can React states be used instead?

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
```diff
@@ -0,0 +1,366 @@
...
+
+import { Messages } from '../../../properties/Messages';
```
vkrizan, line 23: @josejulio General observation: can properties be put on an import path? It could ease future refactoring should this be moved to a different path (or package).

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
```diff
@@ -0,0 +1,366 @@
...
+                        <FlexItem>
+                            <Form>
+                                <FormGroup
+                                    label="Server hostname/IP Address and port (hostname:port)"
+                                    labelIcon={<Popover
+                                        headerContent={<div>
+                                            The server <b>hostname/IP Address</b> and <b>port</b> of your splunk HTTP Event Collector
+                                        </div>}
+                                        bodyContent={<div>
+                                            Usually the port is 8088
```
vkrizan, line 295: `s/Usually the port is/The default port is/`?

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
```diff
@@ -0,0 +1,366 @@
...
+                                            See {' '}
+                                            <a
+                                                target='_blank'
+                                                rel='noreferrer'
+                                                href='https://docs.splunk.com/Documentation/Splunk/8.2.5/Data/UsetheHTTPEventCollector'>
```
vkrizan, line 300: Splunk version within the URL (`8.2.5`) can be replaced with `latest`.

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
```diff
@@ -0,0 +1,366 @@
...
+7) [PUT] Update eventType under /api/notifications/v1.0/notifications/eventTypes/{EVENT_TYPE_UUID}/behaviorGroups with the payload:
+
+["ff59b502-da25-4297-bd88-6934ad0e0d63"] <<- Behavior group ID
+*/
+
+export const SplunkSetupPage: React.FunctionComponent = () => {
+
+    const [ hecId, setHecId ] = useState('');
+    const [ splunkServerHostName, setHostName ] = useState('');
+    const [ automationLogs, setAutomationLogs ] = useState(`CLICK THE BUTTON TO START THE AUTOMATION\n`);
```
vkrizan, line 90: Can the message be cleared right after the automation is started?

vkrizan (commented):
@jpramos123 very neat!

This is no longer a log component, but a whole Splunk setup automation. Great work!

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
josejulio: We could, yes. I guess we can do that in a separate PR.

src/pages/Integrations/SplunkSetup/Constants.ts
```diff
@@ -0,0 +1,23 @@
+export const RBAC_GROUPS_API = '/api/rbac/v1/groups/';
+export const RBAC_ROLES_API = '/api/rbac/v1/groups/{}/roles/';
+export const RBAC_ADD_USER_API = '/api/rbac/v1/groups/{}/principals/';
```
josejulio, line 3: I'm going to add some RBAC API files on a separate PR - i'll let you know when they are in

src/pages/Integrations/SplunkSetup/Constants.ts
```diff
@@ -0,0 +1,23 @@
...
+export const CREATE_INTEGRATION_API = '/api/integrations/v1.0/endpoints';
+export const CREATE_BEHAVIOR_GROUP_API = '/api/notifications/v1.0/notifications/behaviorGroups';
+export const UPDATE_BEHAVIOR_GROUP_API = '/api/notifications/v1.0/notifications/behaviorGroups/{}/actions';
+export const EVENTTYPE_API = '/api/notifications/v1.0/notifications/eventTypes/{}/behaviorGroups';
```
josejulio, line 7: We alredy have these in a typesafe API
Could you take a look at the [fetching library](https://marcin-piela.github.io/react-fetching-library
) we use? 

- [Create Integration](https://github.com/RedHatInsights/notifications-frontend/blob/master/src/services/useSaveIntegration.ts#L37)
- [Create/Update behavior group](https://github.com/RedHatInsights/notifications-frontend/blob/master/src/services/Notifications/SaveBehaviorGroup.ts)
- [Events](https://github.com/RedHatInsights/notifications-frontend/blob/master/src/services/EventLog/GetNotificationEvents.ts)

Let me know if you have questions

src/pages/Integrations/SplunkSetup/Constants.ts
```diff
@@ -0,0 +1,23 @@
...
+
+// If prod use prod UUID, else use stage
+export const NOTIF_ADM_ROLE_UUID = window.insights?.chrome.getEnvironment() === 'prod' ? '' : '15f910ad-150a-46dd-b666-0b21088e9c55';
+export const BEHAVIOR_GROUP_BUNDLE_ID =  window.insights?.chrome.getEnvironment() === 'prod' ? '' : '35fd787b-a345-4fe8-a135-7773de15905e';
```
josejulio, line 11: Needs to be replaced with API calls - this is a specific bundle and admin role?

src/pages/Integrations/SplunkSetup/Constants.ts
josejulio: technically possible, yes, we provide APIs for that. I'm wondering what's the plan. Are you going to provide a wizard to pick the bundle/events/apps that you want to connect or are always the same?

src/pages/Integrations/SplunkSetup/Constants.ts
```diff
@@ -0,0 +1,23 @@
...
+export const SPLUNK_INTEGRATION_NAME = 'SPLUNK_AUTOMATION';
+export const SPLUNK_BEHAVIOR_GROUP_NAME = 'SPLUNK_AUTOMATION_GROUP';
+
+const EVENT_TYPES_PROD = [];
+const EVENT_TYPES_STAGE = [ '6d4d1e03-f1bf-427c-8c62-08a114242b5d', // policy-triggered
+    '95a6d179-5dbd-4de4-a726-5bab2a699912',  // new-recommendation
+    '5df33474-9cfd-40e9-9a86-9f8857147f89', // drift-baseline-detected
+    '75233a10-646c-4e26-a5f1-639a7df8f29f' ]; // resolved-recommendation
+
+export const EVENT_TYPES_IDS_ARR = window.insights?.chrome.getEnvironment() === 'prod' ? EVENT_TYPES_PROD : EVENT_TYPES_STAGE;
```
josejulio, line 23: As a side note, I preffer if we keep the `getEnvironment()` calls inside the components - that way I can mock these values (like i do with the experimental flag) 
You can use `import { getInsights } from '@redhat-cloud-services/insights-common-typescript';` for getting an insights object.

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
josejulio: FWIW we don't use redux here except for connecting with the toasts-api from chrome.

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
```diff
@@ -0,0 +1,366 @@
...
+                                            See {' '}
+                                            <a
+                                                target='_blank'
+                                                rel='noreferrer'
+                                                href='https://docs.splunk.com/Documentation/Splunk/8.2.5/Data/UsetheHTTPEventCollector'>
+                                                Splunk HEC setup
+                                            </a>
+                                        </div>}
+                                    >
+                                        <button
```
josejulio, line 305: Could you use the [patternfly](https://patternfly-react.surge.sh/components/button) button?

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
vkrizan: Of course. Thanks!

src/pages/Integrations/SplunkSetup/Constants.ts
jpramos123: As far I investigated, those UUIDs are the same depending on the environment (stage/prod). The idea was to have those as constants, like this: https://github.com/RedHatInsights/insights-dashboard/blob/master/src/AppConstants.js#L82

src/pages/Integrations/SplunkSetup/Constants.ts
jpramos123: Yes, Notifications Administrator and the bundle is RHEL

src/pages/Integrations/SplunkSetup/Constants.ts
vkrizan: @jpramos123 I would not count on these to be constant across environments, and I'd definitely avoid hardcoding it. We can hardcode the event types for now.

@josejulio Yes, having an option to select which event type triplets to use is on our future radar. For now (for beta) we would pick some, or maybe leverage wildcards.

vkrizan:
I'm going to take over this effort. I'll be updating this PR.

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
vkrizan: Removing...

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
vkrizan: Replacing...

src/pages/Integrations/SplunkSetup/SplunkSetupPage.tsx
vkrizan: Will do...

src/pages/Integrations/SplunkSetup/Constants.ts
vkrizan: The way it is written is more functional. The library heavily uses hooks to make queries which need to be bound to a component on a top level. I'm looking at a way to migrate to it, but if you @josejulio would have an idea I'd love to hear it.

josejulio: You can stick to current approach using [useClient](https://marcin-piela.github.io/react-fetching-library/#/?id=useclient).
Basically it's a replacement for the `axios.post`. See an [example](https://github.com/RedHatInsights/notifications-frontend/blob/4c34bd67589ff678c12479b5c5a84a7e9e80423e/src/app/rbac/RbacGroupContextProvider.tsx#L44).

josejulio:
Closing, as it was replaced by #269

PR #265, opened 2022-03-31
NOTIF-347 Adding rbac groups as recipients of emails

josejulio:
Accepts RBAC groups as email recipients.

Peek 2022-03-31 15-23

Peek 2022-04-12 10-35
Peek 2022-04-12 10-34

### DIFF
--- a/src/generated/OpenapiIntegrations.ts
+++ b/src/generated/OpenapiIntegrations.ts
@@ -12,8 +12,28 @@ import { Action } from 'react-fetching-library';
 import * as z from 'zod';
 
 export namespace Schemas {
+  export const AddAccessRequest = zodSchemaAddAccessRequest();
+  export type AddAccessRequest = {
+    application_id?: UUID | undefined | null;
+    role?: string | undefined | null;
+  };
+
+  export const AddApplicationRequest = zodSchemaAddApplicationRequest();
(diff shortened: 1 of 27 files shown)


josejulio:
I still need to add some tests, but I think we can start to review it.
@mmenestr @jeromemarc

mmenestr:
LGTM from the screen recording!

src/generated/OpenapiIntegrations.ts
```diff
@@ -12,8 +12,28 @@ import { Action } from 'react-fetching-library';
...
+    owner_role?: string | undefined | null;
+  };
+
   export const Application = zodSchemaApplication();
   export type Application = {
+    displayName?: string | undefined | null;
+    id?: UUID | undefined | null;
+  };
+
+  export const Application1 = zodSchemaApplication1();
```
aferd, line 35: just my curiosity as to why this is marked Application1? 

other than that, lgtm!

aferd (approved):

src/generated/OpenapiIntegrations.ts
josejulio: The openapi has the `Application1` - i guess it's because we are using other `Application` type that is not the main one.

PR #269, opened 2022-04-08
[EVNT-358] Splunk Automation Setup Page

vkrizan:
Creates a new page for automated Splunk integration set up.

The automation:
* creates new Integration
* creates new Behavior Group for `rhel` bundle
* Links the integration and behavior group (by adding it as an action)
* Adds an action to selected event types within rhel bundle (Insights)

Screenshot 2022-04-12 at 19-36-07 Integrations Settings

Besides that adds and/or refactors services:
* creates `useGetBundleByName`
* makes `bundleName` of `BehaviorGroup` optional
* creates `useGetAllEventTypes`
* create `useGetAnyBehaviorGroupByNotification`

Co-authored by @jpramos123
Replaces #264

### DIFF
--- a/src/pages/Integrations/SplunkSetup/useSplunkSetup.tsx
+++ b/src/pages/Integrations/SplunkSetup/useSplunkSetup.tsx
@@ -1,0 +1,224 @@
+/*
+Steps:
+
+1) [POST] Create Integrations under /api/integrations/v1.0/endpoints with the payload:
+{
+    "name": "Splunk Automation",
+    "enabled": true,
+    "type": "camel",
+    "sub_type": "splunk",
+    "description": "",
+    "properties": {
+        "url": "http://decd-187-3-186-244.ngrok.io",
+        "disable_ssl_verification": false,
+        "secret_token": "MYHEC_TOKEN",
+        "basic_authentication": {},
+        "extras": {}
+    }
+}
+
+2) Create behavior group under /api/notifications/v1.0/notifications/behaviorGroups with the payload:
+
+{
+  "bundle_id":"35fd787b-a345-4fe8-a135-7773de15905e",
+  "display_name":"Splunk-automation"
+}
+
+3) [POST] Update behavior group under api/notifications/v1.0/notifications/behaviorGroups/{BEHAVIOR_GROUP_ID}/actions with the payload:
+
+  ["8d8dca57-1834-48dd-b6ac-265c949c5e60"] <<-- Id of the integration
+
+4) [PUT] Update eventType under /api/notifications/v1.0/notifications/eventTypes/{EVENT_TYPE_UUID}/behaviorGroups with the payload:
+
+["ff59b502-da25-4297-bd88-6934ad0e0d63"] <<- Behavior group ID
+*/
+
+import { useClient } from 'react-fetching-library';
+
+import { useGetAllEventTypes } from '../../../services/GetEventTypes';
+import { useGetAnyBehaviorGroupByNotification } from '../../../services/Notifications/GetBehaviorGroupByNotificationId';
+import { useGetBundleByName } from '../../../services/Notifications/GetBundles';
+import { linkBehaviorGroupAction } from '../../../services/Notifications/LinkBehaviorGroup';
+import { useSaveBehaviorGroupMutation } from '../../../services/Notifications/SaveBehaviorGroup';
+import { useUpdateBehaviorGroupActionsMutation } from '../../../services/Notifications/UpdateBehaviorGroupActions';
+import { useSaveIntegrationMutation } from '../../../services/useSaveIntegration';
+import { Integration, IntegrationCamel, IntegrationType, NewIntegrationTemplate } from '../../../types/Integration';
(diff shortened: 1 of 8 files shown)


src/pages/Integrations/SplunkSetup/useSplunkSetup.tsx
```diff
@@ -0,0 +1,162 @@
...
+        onProgress('Creating Splunk Integration...');
+        const integration = await createSplunkIntegration({ integrationName, hecToken, splunkServerHostName });
+        onProgress(`  ${integrationName} integration created.`);
+
+        onProgress('Creating Splunk Behavior Group...');
+        const behaviorGroup = await createSplunkBehaviorGroup({ behaviorGroupName, bundleName });
+        onProgress(`  ${behaviorGroupName} behavior group created.`);
+
+        onProgress('Adding Splunk integration as an action behavior group...');
+        updateSplunkBehaviorActions(behaviorGroup, integration);
```
josejulio: I suppose the `await`  is missing here

vkrizan: Thanks, added that

vkrizan:
@josejulio @jpramos123 ready for review. Note that we'll update the visuals (wizard) separately.

@josejulio Let me know if you'd need to have the services improvements separately. Also would it be possible to not squash it but merge as is?

Thank you.

I'll go through the lint errors tomorrow.

josejulio:
> @josejulio Let me know if you'd need to have the services improvements separately. Also would it be possible to not squash it but merge as is?

We usually `squash and merge` - but I saw that you took the time to split each step in a separate commit. It's fine by me to merge without squash

src/pages/Integrations/SplunkSetup/useSplunkSetup.tsx
```diff
@@ -0,0 +1,207 @@
...
+            isEnabled: true,
+            sslVerificationEnabled: true,
+        };
+
+        const { errorObject, payload } = await mutate(newIntegration);
+        if (errorObject) {
+            throw errorObject;
+        }
+
+        return payload?.value as Integration;
```
josejulio: The payload implements discrimination by `payload?.status` or `payload?.type`.
While is not required to modify this PR, one should be able to do:
```suggestion
        const { payload } = await mutate(newIntegration);
        if (payload?.status !== 200) {
            throw new Error('Invalid response');
        }

        return payload.value;
```

josejulio:
> I'll go through the lint errors tomorrow.

You can start with `yarn lint:ts:fix`

src/pages/Integrations/SplunkSetup/useSplunkSetup.tsx
vkrizan: I'm trusting the library to produce `errorObject` if there is an error. Comparing a single HTTP status codes is not ideal either, as it might be any of the `2XX` responses. In this POST case it should be `201 Created`. Check https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses.

I acknowledge that even the check of errorObject does not cover all cases, i.e. the case when the payload is not correct or passed.

vkrizan: Also check https://marcin-piela.github.io/react-fetching-library/#/?id=error-boundaries

vkrizan: Okay, I've added additional checks for `error`. Those would cover it, when the error object is not thrown. `error` is set for non `2xx` responses.

https://github.com/marcin-piela/react-fetching-library/blob/9b0abd81cf7494931ca144a1d64952f6e0d77b18/src/client/client.ts#L77
https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

josejulio: One of these layers that were added is a validation layer - all the requests are validated (runtime) and discriminated types are created (compile time) to ensure the types are correct. 

It's maybe not ideal, but this allows us to detect api changes at compile time (given the openapi is updated).

Aside the `status` a type is added to the payload, in this case, if the type is `"Integration"` the payload.value is constrained to `Integration` and below cast is not needed anymore.

The implementation of these mechanisms needs to be documented and needs polishing to make it usable and friendly to the implementer, but it's something that is already there. 

Screenshot_2022-04-13_06-50-25

An example on other request, the IDE will even help with hinting the possible types:

Screenshot_2022-04-13_06-57-10

Screenshot_2022-04-13_06-58-51

I'm not requiring these changes on this PR. Current checks are OK  and can be refined later.

josejulio (approved):
lgtm

src/pages/Integrations/SplunkSetup/useSplunkSetup.tsx
vkrizan: Yeah, thanks. It would be great to have this generalized and avoid repeating it for each new api hook.

I've added couple of these notes to https://issues.redhat.com/browse/NOTIF-602

PR #284, opened 2022-04-29
NOTIF-154 Disable actions without integrations

josejulio:
- shows a warning and a link to the integrations page
Peek 2022-04-29 11-28


aferd (approved):
lgtm!

PR #285, opened 2022-04-29
Update to latest generated types

josejulio:
These are mostly generated names changes by the renaming of Services to Resources.


josejulio:
Ah, so I already did that :facepalm: should be the same as  #288

PR #288, opened 2022-05-04
Update schema with recent changes from resources vs service

josejulio:


josejulio:
I'm merging now - some changes in the autogenerated API data.

PR #293, opened 2022-05-16
fix(splunk): form validation of URL and UUID

vkrizan:
Fixes URL validation in the Splunk form to accept dashes and other
characters that can be found within a URL.

Uses yup with customized test for checking paths and queries.

Validates UUID with yup lib.

And lastly improves the form inputs performance.

EVNT-475


vkrizan:
@josejulio please review. Thanks.

josejulio (approved):

PR #297, opened 2022-06-03
NOTIF-537 Add margin to the "View Event Log" button

josejulio:
#### Before
Screenshot_2022-06-03_11-02-16

#### After
Screenshot_2022-06-03_11-02-30


aferd (approved):
lgtm

PR #299, opened 2022-06-07
NOTIF-618 disabled integration should be grey

aferd:
before:
Screen Shot 2022-06-07 at 1 47 04 PM

after:
Screen Shot 2022-06-07 at 2 08 18 PM


josejulio (approved):
lgtm

PR #318, opened 2022-07-11
NOTIF-436 Limit behavior group name length to 150 characters

Mishrasubha:
Screenshot from 2022-07-11 18-15-08

### DIFF
--- a/src/pages/Notifications/BehaviorGroupWizard/Steps/BasicInformationStep.tsx
+++ b/src/pages/Notifications/BehaviorGroupWizard/Steps/BasicInformationStep.tsx
@@ -28,7 +28,7 @@ const BasicInformationStep: React.FunctionComponent = () => {
 };
 
 export const schema = Yup.object({
-    displayName: Yup.string().min(1).required('Behavior group name is required')
+    displayName: Yup.string().min(1).max(150, 'Must be 150 characters or less').required('Behavior group name is required')
 });
 
 export const useBasicInformationStep: CreateWizardStep = () => {


src/pages/Notifications/BehaviorGroupWizard/Steps/BasicInformationStep.tsx
```diff
@@ -28,7 +28,7 @@ const BasicInformationStep: React.FunctionComponent = () => {
 };
 
 export const schema = Yup.object({
-    displayName: Yup.string().min(1).required('Behavior group name is required')
+    displayName: Yup.string().min(1).max(150).required('Behavior group name is required')
```
josejulio: @mmenestr What error should be displayed if the name is over 150 characters?

mmenestr: I would display help text under the input with "Must be 150 characters or less." and if they exceed it, it turns red and the input takes on an error state
image

like this but instead of "invalid input" it says "Must be 150 characters or less." ^

josejulio: @Mishrasubha could you add the message as above?

Should be something like:

```suggestion
    displayName: Yup.string().min(1).max(150, 'Must be 150 characters or less').required('Behavior group name is required')
```

Mishrasubha: On it

josejulio (approved):
lgtm!

mmenestr:
Could we change this to "Must be 150 character or less" instead of "displayName must be at most 150 characters" like was discussed in the other PR?

josejulio:
The screenshot is outdated, with latest changes the message shows as "Must be 150 character or less"

josejulio:
Screenshot_2022-07-12_08-54-12

mmenestr:
Oh ok, thank you so much!

PR #320, opened 2022-07-14
Update openapi generated files

josejulio:


josejulio:
Merging now, removes account_id and org_id fields (internal)

PR #326, opened 2022-08-09
NOTIF-735 Adds service-now integration (experimental)

josejulio:
The Service now integration type can only be accessed behind the experimental switch.
Internally, it is a `camel` type with the subtype `service_now`

Screenshot_2022-08-09_14-06-23

// cc: @vkrizan @jeromemarc @jpramos123

### DIFF
--- a/src/components/Integrations/Form/IntegrationTypeForm.tsx
+++ b/src/components/Integrations/Form/IntegrationTypeForm.tsx
@@ -3,10 +3,10 @@ import { assertNever } from 'assert-never';
 import * as React from 'react';
 
 import { IntegrationType, isCamelType, UserIntegrationType } from '../../../types/Integration';
+import { IntegrationTypeCamelExtrasForm } from './IntegrationTypeCamelExtrasForm';
 import { IntegrationTypeCamelForm } from './IntegrationTypeCamelForm';
 import { IntegrationTypeHttpForm } from './IntegrationTypeHttpForm';
 import { IntegrationTypeSlackForm } from './IntegrationTypeSlackForm';
-import { IntegrationTypeSplunkForm } from './IntegrationTypeSplunkForm';
 
 export interface IntegrationTypeForm extends OuiaComponentProps {
     type: UserIntegrationType;
@@ -17,7 +17,8 @@ export const IntegrationTypeForm: React.FunctionComponent<IntegrationTypeForm> =
     if (isCamelType(props.type)) {
         switch (props.type) {
             case UserIntegrationType.SPLUNK:
-                return <IntegrationTypeSplunkForm { ...props } />;
+            case UserIntegrationType.SERVICE_NOW:
+                return <IntegrationTypeCamelExtrasForm { ...props } />;
             case UserIntegrationType.SLACK:
                 return <IntegrationTypeSlackForm { ...props } />;
         }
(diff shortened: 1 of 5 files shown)


src/types/Integration.ts
```diff
@@ -7,12 +7,14 @@ export enum IntegrationType {
     WEBHOOK = 'webhook',
     EMAIL_SUBSCRIPTION = 'email_subscription',
     SPLUNK = 'camel:splunk',
-    SLACK = 'camel:slack'
+    SLACK = 'camel:slack',
+    SERVICE_NOW = 'camel:service_now'
```
vkrizan: Can it be without the underscore? Either `servicenow` or `snow`.

src/config/Config.ts
```diff
@@ -24,6 +24,9 @@ const integrationTypes: Record<IntegrationType, IntegrationTypeConfigBase> = {
     [IntegrationType.SPLUNK]: {
         name: 'Splunk'
     },
+    [IntegrationType.SERVICE_NOW]: {
+        name: 'Service now'
```
vkrizan: `s/Service now/ServiceNow/`

vkrizan (commented):
Form looks good. I left some name changes recommendations.

src/components/Integrations/Form/IntegrationTypeForm.tsx
```diff
@@ -17,7 +17,8 @@ export const IntegrationTypeForm: React.FunctionComponent<IntegrationTypeForm> =
     if (isCamelType(props.type)) {
         switch (props.type) {
             case UserIntegrationType.SPLUNK:
-                return <IntegrationTypeSplunkForm { ...props } />;
+            case UserIntegrationType.SERVICE_NOW:
+                return <IntegrationTypeCamelExtrasForm { ...props } />;
```
vkrizan, line 21: Actually the form mentions Splunk:
183751060-6ece0ba6-b593-4eed-8b8e-63eb89bc7ff4

The secret token would for ServiceNow be a user password. Maybe we can generalize the wording for ServiceNow.

src/types/Integration.ts
josejulio: Sure, it can be `servicenow` - any particular reason you don't want the underscore? or just matter of preference?

src/config/Config.ts
josejulio: Whoops - didn't know how the name of the service was spelt, thanks!

src/components/Integrations/Form/IntegrationTypeForm.tsx
josejulio: Maybe we can go with `The defined secret token is sent to the integration for authentication purposes depending each integration needs`.

Alternatively  we could use a different wording for each service to make it more specific. 
Do you have a specific for ServiceNow ?

aferd (approved):

josejulio:
Merging now - we need to tweak later if we want to use a different message for `ServiceNow` or refactor the message to make sense in both `Splunk` and `ServiceNow`.

src/components/Integrations/Form/IntegrationTypeForm.tsx
vkrizan: No specifics at the moment for SNow. It might be a "Password of a ServiceNow integration user.".

Can we differentiate the text based on the type?

josejulio: yes, we can do that.

PR #327, opened 2022-08-10
NOTIF-747 Modifies Behavior group list scroll direction (horizontally…

josejulio:
… now)

Peek 2022-08-10 06-56


josejulio:
@jeromemarc and @mmenestr could you take a look?

aferd (approved):

mmenestr:
that looks great!! Can the user also horizontal scroll using a track pad or mouse? ie, could they scroll without having to click?

josejulio:
> that looks great!! Can the user also horizontal scroll using a track pad or mouse? ie, could they scroll without having to click?

That's a native scroll - I didn't test, but should behave like any other horizontal scroll in the computer.

PR #334, opened 2022-09-08
NOTIF-771 Updating ServiceNow token description

josejulio:
Screenshot_2022-09-08_08-26-16
Screenshot_2022-09-08_08-26-01

@jeromemarc @vkrizan


vkrizan (approved):
LGTM

PR #335, opened 2022-09-08
Expose ServiceNow to beta

josejulio:
@vkrizan @jeromemarc 
It's ok to release this already to prod-beta?

If not, let me know when it will be a good time.


vkrizan:
It is fine by me. It would help QE as well.

PR #345, opened 2022-10-03
NOTIF-802 Display the endpoint status

josejulio:
Adding integrations status

The integration status is used for camel integrations that require a setup phase - so the flow will be:

Screenshot_2022-10-03_12-32-11

The status is aggregated with the result of the last notification histories if the status is READY - otherwise it will display the status

The header contains a `?` button with the description of the status

Screenshot_2022-10-18_09-11-13

####  No history - shows as Ready

Screenshot_2022-10-17_17-39-25

#### With history - last history is success but has errors
Screenshot_2022-10-17_17-38-58

Note: When there are only success calls, it does not show the "Degraded connection"

#### With history - last history is error
Screenshot_2022-10-17_17-37-38

### Provisioning or Deleting
Screenshot_2022-10-17_17-40-36

### Failed to create
Screenshot_2022-10-17_17-39-54

Notes: 
 - This PR does not move the last errors out of the expanded content
 -  Some screenshots don't have the `?` in the header of the table - these were taken before adding it.

### DIFF
--- a/src/components/Integrations/Table.tsx
+++ b/src/components/Integrations/Table.tsx
@@ -1,4 +1,4 @@
-import { Button, ButtonVariant, EmptyStateVariant, Popover, Spinner, Switch } from '@patternfly/react-core';
+import { Button, ButtonVariant, EmptyStateVariant, Spinner, Switch } from '@patternfly/react-core';
 import { CubesIcon, HelpIcon } from '@patternfly/react-icons';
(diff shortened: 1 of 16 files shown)


josejulio:
@jeromemarc @mmenestr Could you review the UI for the endpoint status?

Thanks!

jeromemarc:
@josejulio Looks good!

PR #363, opened 2022-11-15
Migration to FEC

JuliePrazakova:


josejulio:
Looks like there are conflicting type definitions for react - forcing (adding to `package.json`) to use the top one should fix that compile issue.

```json
"overrides": {
     "@types/react": "17.0.39"
}
```

PR #376, opened 2023-01-30
RHCLOUD-23476 expose daily email timezone config in settings > notif

aferd:


src/services/Notifications/SaveTimePreference.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { Operations, Schemas } from '../../generated/OpenapiNotifications';
+
+type SaveTimePrefActionsParams = {
+    body: Schemas.LocalTime;
+}
+
+const updateTimePrefActionCreator = (params: SaveTimePrefActionsParams) => {
+    return Operations.OrgConfigResourceSaveDailyDigestTimePreference.actionCreator({
+        body: params.body
```
aferd, line 11: I have a question here, wouldn't I need a bundleId param to be able to save it to the specific bundle? Because right now, when I input a time it is saving for all bundles...

josejulio: It's global to the account, not specific to a particular bundle so i think it's fine.

aferd:
https://user-images.githubusercontent.com/86322239/220739117-cb7496b2-e997-479d-8e99-d6a7bf2a4dfb.mov

aferd:
tests are failing due to what looks like changes from updating the generated files.

@josejulio do you know why we got rid of the "default" property?

aferd:
https://user-images.githubusercontent.com/86322239/223459299-02d4260b-bf7c-4de0-b062-c5b97d4d0f3c.mov

works as expected, errors shown are mentioned in above comment.

josejulio:
> @josejulio do you know why we got rid of the "default" property?

Yes, it was deprecated and no longer used. We ensured that no one is still using that and removed the types. 

We could remove it from the frontend as well.  Either in this PR or a separate one.

aferd:
> > @josejulio do you know why we got rid of the "default" property?
> 
> Yes, it was deprecated and no longer used. We ensured that no one is still using that and removed the types.
> 
> We could remove it from the frontend as well. Either in this PR or a separate one.

got it, i will remove this in a separate PR.

maryshak1996:
Hey @aferd! just a UX review of this and overall the spacing and all looks spot on :) a couple suggestions though
- Can we make it such that the time and time zone fields only appear once someone clicks on 'Custom time'? And add a bit of helper text below to indicate that that would happen? At initial glance, I assumed that everyone had to fill out those fields even if they selected 'Default time', and this would alleviate that issue.
- I noticed a slight difference in in the Time zone selector you have and what patternfly suggests (here is the [PF time zone guidelines](https://www.patternfly.org/v4/components/time-picker/design-guidelines#time-picker-with-time-zones)) -- see my image below for how to make it more PF consistent
 Daily-digest-form_March9-update

aferd:
https://user-images.githubusercontent.com/86322239/229571128-5d34646e-0cb1-453d-88d4-7fcf71add166.mov

src/components/Notifications/TimeConfig.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    const [ radioSelect, setRadioSelect ] = React.useState(false);
+    const [ showCustomSelect, setShowCustomSelect ] = React.useState(false);
+
+    const saveTimePreference = useUpdateTimePreference();
+
+    const handleRadioSelect = React.useCallback(() => {
+        setRadioSelect(true);
+        setShowCustomSelect(false);
+
```
josejulio:
```suggestion
```

src/components/Notifications/TimeConfig.tsx
```diff
@@ -0,0 +1,115 @@
...
+        setShowCustomSelect(true);
+    }, []);
+
+    const handleTimeSelect = React.useCallback((time) => {
+        setIsOpen(false);
+        const mutate = saveTimePreference.mutate;
+        mutate({
+            body: time
+        });
+
```
josejulio:
```suggestion
```

src/components/Notifications/TabComponent.tsx
```diff
@@ -0,0 +1,39 @@
...
+    return (
+        <div className={ backgroundColorClassName }>
+            <Tabs className={ paddingLeftClassName }
+                defaultActiveKey={ activeTabKey } role="region" onClick={ handleTabClick }>
+                <Tab eventKey={ 0 } title={ <TabTitleText>Configuration</TabTitleText> }>
+                    { props.elements[0]}
+                </Tab>
+                <Tab eventKey={ 1 } title={ <TabTitleText>Settings</TabTitleText> }>
+                    { props.elements[1] }
+                </Tab>
```
josejulio: I think it will be more clear if you extract this out of this component. The order would be really easy to get mixed.

Step1 (you will need to clear imports):

```suggestion
                {props.children}
```

or instead you could use "settings" and "configuration" props:

```
interface MyTabComponentProps {
    configuration: React.ReactNode;
    settings: React.ReactNode;
  }

```

src/pages/Notifications/List/BundlePage.tsx
```diff
@@ -26,22 +28,26 @@ export const NotificationListBundlePage: React.FunctionComponent<NotificationLis
...
+
+        <TabComponent elements={ [
+            <Main key='0'>
                 <SplunkBetaEnvironmentBanner />
                 <BundlePageBehaviorGroupContent applications={ props.applications } bundle={ props.bundle } />
+            </Main>,
+            <Main key='1'>
+                <TimeConfigComponent />
             </Main>
+        ] } />
```
josejulio: Step 2 (will require fix imports)

```suggestion
        <TabComponent>
            <Tab eventKey={ 0 } title={ <TabTitleText>Configuration</TabTitleText> }>
                 <Main>
                     <SplunkBetaEnvironmentBanner />
                     <BundlePageBehaviorGroupContent applications={ props.applications } bundle={ props.bundle } />
                 </Main>
            </Tab>
            <Tab eventKey={ 1 } title={ <TabTitleText>Settings</TabTitleText> }>
                <Main>
                    <TimeConfigComponent />
                </Main>
            </Tab>
        </TabComponent>
```

josejulio (changes requested):
Some minor details to make it more clear the tabbed contents.

I also see that we are missing a way to load the current values - The best would be to make that change in a separate PR - do we even have that service available?

src/components/Notifications/TimeConfig.tsx
aferd: It is available, I will make a separate PR for adding this service.

josejulio (approved):
lgtm

PR #385, opened 2023-03-24
RHCLOUD-24868 Fix Teams integration type name

gwenneg:


src/config/Config.ts
```diff
@@ -37,7 +37,7 @@ const integrationTypes: Record<IntegrationType, IntegrationTypeConfigBase> = {
         name: 'Email'
     },
     [IntegrationType.TEAMS]: {
-        name: 'Teams'
+        name: 'Microsoft Teams'
```
gwenneg, line 40: @josejulio Is this the good place to fix so that the frontend shows `Microsoft Teams` instead of `Teams` while creating an integration?

josejulio: Yes, this is a display only string.

josejulio (approved):

PR #388, opened 2023-03-29
RHCLOUD-24987 Hide non used data on special environments

josejulio:
- Fix actions displayed on each environment
 - Fix: Splunk/SNow banner should only be displayed on stage/prod stable


aferd:
looks like there is a lint error but once that is fixed - 

lgtm!

aferd (approved):

PR #392, opened 2023-04-06
RHCLOUD-23554 Add text in UI to inform user that an email notif is being forced

aferd:
Screen Shot 2023-04-06 at 1 23 35 PM
this is what it would look like when there is a forced email. there is currently NO apps sending forced emails so you will not see this in the UI as of now.

### DIFF
--- a/src/components/Notifications/Recipient.tsx
+++ b/src/components/Notifications/Recipient.tsx
@@ -5,7 +5,7 @@ import { join } from '@redhat-cloud-services/insights-common-typescript';
 import * as React from 'react';
 import { style } from 'typestyle';
 
-import { Action, EmailSystemProperties, NotificationType } from '../../types/Notification';
+import { Action, NotificationType } from '../../types/Notification';
 import { NotificationRbacGroupRecipient, NotificationUserRecipient } from '../../types/Recipient';
 import { GroupNotFound } from './Rbac/GroupNotFound';
 
@@ -27,8 +27,6 @@ const greyColorName = style({
 const CommaSeparator = () => <span>, </span>;
 
 export const Recipient: React.FunctionComponent<RecipientProps> = (props) => {
-
-    const [ ignorePref ] = React.useState<EmailSystemProperties>();
 
     if (props.action.type === NotificationType.INTEGRATION) {
         return (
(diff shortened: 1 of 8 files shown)


src/components/Notifications/Recipient.tsx
```diff
@@ -27,6 +27,9 @@ const greyColorName = style({
 const CommaSeparator = () => <span>, </span>;
 
 export const Recipient: React.FunctionComponent<RecipientProps> = (props) => {
+
+    const [ ignorePref ] = React.useState<EmailSystemProperties>();
```
josejulio: I think you still need to load this value into the `Action`.

Here is how to do it:

1. Update [NotificationAdapter](https://github.com/josejulio/notifications-frontend/blob/f711c5d03c6f66372489b3b44c9355cbc1b2fc5b/src/types/adapters/NotificationAdapter.ts#L35) to use the ignore preferences variable (`integration.ignorePreferences`) and pass it to `NotificationUserRecipient`
2. Update [NotificationUserRecipient](https://github.com/josejulio/notifications-frontend/blob/ffc76263cfc72f160bc31f693c3ffbeaa5d395b2/src/types/Recipient.ts#L56) to receive a third boolean for ignorePreferences, and assign it accordingly. You will need to update other places where it's called.
3.  The setting is per recipient. i.e. the recipient "all users" might not be forced but "send to admins" might be forced so I suggest to update:
`Users: {join(users.map(u => u.displayName), CommaSeparator)}` to something like:
```
Users: {join(users.map(u => <>
  {u.displayName}
  {u.ignoreUserPreferences && <span className={ greyColorName }> { props.action.integration.name }
                        <Tooltip content="This integration has been disabled. This action wil not fire until it is enabled." position="bottom">
                            <BanIcon className={ disabledLabelClassName } />
                        </Tooltip>
                    </span>)
</>), CommaSeparator)}
```
The lock icon should show only for the users that are forced.

Alternatively you could display it on the end by checking if there is at least one forced recipient:
```
const ignoreUserPreferences = users.some(u => u.ignoreUserPreferences)
```

4. Theorically the groups could also be forced, but there is no way to set it right now.

josejulio (changes requested):

aferd:
@josejulio when you get time can you re-review? thanks!

src/types/Recipient.ts
```diff
@@ -72,6 +74,7 @@ export class NotificationUserRecipient extends BaseNotificationRecipient {
         );
 
         this.sendToAdmin = sendToAdmin;
+        this.ignorePreferences = ignorePreferences;
     }
 
     public equals(recipient: Recipient) {
```
josejulio, line 80: Could you update this equal to include the `ignorePreferences`? 

Change `return recipient.sendToAdmin === this.sendToAdmin;` to `return recipient.sendToAdmin === this.sendToAdmin && recipient.ignorePreferences === this.ignorePreferences;`

josejulio (approved):
lgtm

PR #397, opened 2023-04-19
RHCLOUD-25364 UX updates to timezone config in Notif

aferd:
UX updates..
1. Change this into a vertical layout instead of side by side, so that the options are below the header and not to the right of it.
2. Make sure the spacing everywhere follows the spacers in the mock. The card inside the body of the page should only be 24px away from the edges, not as far away as it is currently
3. Add Save button to save the changes- and show a toast alert when someone presses the change to confirm it worked.
4. Update the default time to 00:00 UTC, not 12:00 UTC (that was an error on my part, sorry!) 
5. When the user doesn't have the custom time selected, disable the options. Or hide it if that's already been implemented... It's just so that the save button doesn't look super far away if the default is selected.(_this was already done in previous PR_)


aferd:
https://user-images.githubusercontent.com/86322239/235964346-3570b787-0f02-4d11-853c-d7081350ca57.mov

josejulio (approved):

PR #398, opened 2023-04-21
RHCLOUD-25371 add get time pref service

aferd:


src/components/Notifications/TimeConfig.tsx
```diff
@@ -15,8 +16,17 @@ export const TimeConfigComponent: React.FunctionComponent = () => {
...
 
+    const getTimePreference = useGetTimePreference();
     const saveTimePreference = useUpdateTimePreference();
 
+    const timePref = useMemo(() => {
+        if (getTimePreference.payload?.status === 200) {
+            return getTimePreference.payload.value;
+        }
+
+        return undefined;
```
josejulio, line 27: Could we add some sort of loading indicator when we get undefined? Maybe replace the components with a Skeleton https://patternfly-react.surge.sh/components/skeleton

aferd: I will be adding toast notifications to the page on success/failure with this [ticket](https://issues.redhat.com/browse/RHCLOUD-25364). With that, do you also want a skeleton?

src/components/Notifications/TimeConfig.tsx
```diff
@@ -1,22 +1,36 @@
...
+    const getTimePreference = useGetTimePreference();
     const saveTimePreference = useUpdateTimePreference();
 
+    const timePref = useMemo(() => {
+        if (getTimePreference.payload?.status === 200) {
+            return getTimePreference.payload.value;
+        }
+
+        if (getTimePreference.payload?.status === undefined) {
+            return timePrefSkeleton && undefined;
```
josejulio: :thinking: Why do you need to do the `&& undefined` ? 

I think you could simply this with only:

```suggestion
            return <Skeleton />;
```

and by removing the other variable.

josejulio: If the issue is that some of the other components require a string, you could simply return undefined here and do elsewhere:

`timePref ? <TheOtherComponent> : <Skeleton />`

josejulio (approved):

PR #402, opened 2023-04-26
Migration to FEC

josejulio:
Trying something to migrate to FEC


aferd (approved):

Hyperkid123 (commented):
A couple of things.

You want to upgrade to the latest version of the `@redhat-cloud-services/frontend-components-config` package. (Should be 4.7.3).

You should also add `.cache` directory to `.gitignore`. You also want to remove webpack dependencies (except for the bundle analyzer) and possibly some `babel` dependencies as they are actually dependencies of the config package.

josejulio:
> You want to upgrade to the latest version of the `@redhat-cloud-services/frontend-components-config` package. (Should be 4.7.3).

Missed that, thanks!


> You should also add `.cache` directory to `.gitignore`. 

+1

> You also want to remove webpack dependencies (except for the bundle analyzer) and possibly some `babel` dependencies as they are actually dependencies of the config package.

Yes, most of the errors were caused by dependencies being duplicated - I'll try to prune a bit more of the deps.

Hyperkid123 (approved):

PR #414, opened 2023-05-04
Fix preview redirects from within the app itself

karelhala:
### Description

When we migrated to preview we've forgot to update the [insights-common-typescript](https://github.com/RedHatInsights/insights-common-typescript) package and fix the tests. This PR updates both of these things

### JIRA

RHCLOUD-25749


josejulio (approved):

src/utils/__tests__/Basename.test.ts
```diff
@@ -6,14 +6,14 @@ describe('src/utils/Basename', () => {
     });
 
     it('If first is beta, also get the next', () => {
-        expect(getBaseName('/beta/foo/bar/baz')).toEqual('/beta/foo');
+        expect(getBaseName('/preview/foo/bar/baz')).toEqual('/preview/foo');
     });
 
     it('gets the subapp', () => {
-        expect(getSubApp('/foo/bar/baz')).toEqual('foo');
+        expect(getSubApp('/preview/bar/baz')).toEqual('foo');
```
josejulio: If something other than the beta/preview is used, it retrieves the base

```suggestion
        expect(getSubApp('/foo/bar/baz')).toEqual('foo');
```

PR #428, opened 2023-06-16
UX updates to timezone config

aferd:
A few UX updates from Margot...
1. Default time should be selected as "default" setting
2. Default time should be 00:00 UTC midnight
3. Align dropdowns with custom time vs. radio button

### DIFF
--- a/src/components/Notifications/TimeConfig.tsx
+++ b/src/components/Notifications/TimeConfig.tsx
@@ -91,7 +91,6 @@ export const TimeConfigComponent: React.FunctionComponent = () => {
                         <br></br>
                         <Split>
                             <SplitItem isFilled>
-                                {/* {timePref ? */}
                                 <Stack hasGutter>
                                     <StackItem>
                                         <Radio
--- a/src/components/Notifications/TimeConfig.tsx
+++ b/src/components/Notifications/TimeConfig.tsx
@@ -35,6 +35,7 @@ export const TimeConfigComponent: React.FunctionComponent = () => {
         return undefined;
 
     }, [ getTimePreference.payload?.status, getTimePreference.payload?.value ]);
+    console.log(getTimePreference);
 
     const handleRadioSelect = React.useCallback(() => {
         setRadioSelect(true);
@@ -56,10 +57,10 @@ export const TimeConfigComponent: React.FunctionComponent = () => {
         setIsOpen(false);
     }, [ timeSelect ]);
 
-    const handleButtonSave = React.useCallback((timeSelect) => {
+    const handleButtonSave = React.useCallback(() => {
         const mutate = saveTimePreference.mutate;
         mutate({
-            body: timeSelect.body
+            body
         }).then((response) => {
             if (response.status === 200) {
                 return (
--- a/src/components/Notifications/TimeConfig.tsx
+++ b/src/components/Notifications/TimeConfig.tsx
@@ -1,7 +1,7 @@
-
 import { Alert, Button, Card, CardBody, CardFooter, Dropdown, DropdownItem, DropdownToggle, HelperText, HelperTextItem,
-    Radio, Skeleton, Split, SplitItem, Stack, StackItem,
+    Radio, Split, SplitItem, Stack, StackItem,
     Text, TextVariants, TimePicker, Title } from '@patternfly/react-core';
+import { global_spacer_lg } from '@patternfly/react-tokens';
 import React, { useMemo } from 'react';
 import timezones from 'timezones.json';
 import { style } from 'typestyle';
@@ -13,9 +13,13 @@ const dropDownClassName = style({
     width: '280px'
 });
 
+const dropDownPaddingClassName = style({
+    paddingLeft: global_spacer_lg.value
+});
+
 export const TimeConfigComponent: React.FunctionComponent = () => {
 
-    const [ radioSelect, setRadioSelect ] = React.useState(false);
+    const [ radioSelect, setRadioSelect ] = React.useState(true);
     const [ showCustomSelect, setShowCustomSelect ] = React.useState(false);
 
     const getTimePreference = useGetTimePreference();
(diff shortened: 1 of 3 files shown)


aferd:
@josejulio the "save" is not getting hit and I think this is the last piece of this that needs work - can you review when you get a chance?

aferd:
Screen Shot 2023-06-16 at 3 52 24 PM
Screen Shot 2023-06-16 at 3 53 10 PM

src/components/Notifications/TimeConfig.tsx
```diff
@@ -73,65 +77,65 @@ export const TimeConfigComponent: React.FunctionComponent = () => {
...
-                                            <Radio
-                                                isChecked={ radioSelect && !showCustomSelect }
-                                                onChange={ handleRadioSelect }
-                                                value={ timePref }
-                                                id='settings-time-config'
-                                                label='Default time'
-                                                description='12:00 UTC'
-                                                name='radio-select'>
-                                            </Radio>
+                                {/* {timePref ? */}
```
josejulio: Is this supposed to be removed?

src/components/Notifications/TimeConfig.tsx
```diff
@@ -73,65 +77,65 @@ export const TimeConfigComponent: React.FunctionComponent = () => {
...
-                                    : <Skeleton />}
+                                                </DropdownToggle> }
+                                                isOpen={ isOpen }
+                                                onSelect={ handleTimeSelect }
+                                                menuAppendTo={ () => document.body }
+                                                dropdownItems={ dropdownItems }>
+                                            </Dropdown>
+                                        </StackItem></>)}
+                                </Stack>
+                                {/* : <Skeleton />} */}
```
josejulio: Is this supposed to be removed?

josejulio:
Missed this from other review, but https://github.com/RedHatInsights/notifications-frontend/blob/master/src/services/Notifications/GetTimePreference.ts#LL14C12-L14C23 should be `useQuery` instead of `useMutate`

josejulio:
https://github.com/RedHatInsights/notifications-frontend/blob/master/src/components/Notifications/TimeConfig.tsx#L52

Here, the function receives `time`, but is not what you expect. The argument comes from the `onClick` function which is usually an event. 

I think you would need to use `handleTimeSelect` to store in a state the selected time and use that value in the function.

src/components/Notifications/TimeConfig.tsx
```diff
@@ -45,14 +51,15 @@ export const TimeConfigComponent: React.FunctionComponent = () => {
...
 
-    const handleTimeSelect = React.useCallback(() => {
+    const handleTimePrefSelect = React.useCallback(() => {
+        setTimeSelect(timeSelect);
         setIsOpen(false);
-    }, []);
+    }, [ timeSelect ]);
 
-    const handleButtonSave = React.useCallback((time) => {
+    const handleButtonSave = React.useCallback((timeSelect) => {
```
josejulio:
```suggestion
    const handleButtonSave = React.useCallback(() => {
```

Remove `timeSelect`, it's not a parameter sent to this function. This function only receives an event.
You want to use `timeSelect` from the outer scope.

josejulio: You will need to add `timeSelect` to the dependencies of ths `useCallback` function:
https://github.com/RedHatInsights/notifications-frontend/blob/f75e16d225d4ce0b9759f818aab8678bf3f1ddf3/src/components/Notifications/TimeConfig.tsx#L74

src/components/Notifications/TimeConfig.tsx
```diff
@@ -73,71 +90,67 @@ export const TimeConfigComponent: React.FunctionComponent = () => {
...
+                                            label='Default time'
+                                            description='00:00 UTC'
+                                            name='radio-select'>
+                                        </Radio>
+                                    </StackItem>
+                                    <StackItem>
+                                        <Radio
+                                            isChecked={ radioSelect && showCustomSelect }
+                                            onChange={ handleCustomRadioSelect }
+                                            id='settings-time-config'
```
josejulio: We need to use a different id here.
```suggestion
                                            id='settings-time-config-custom'
```

src/components/Notifications/TimeConfig.tsx
```diff
@@ -1,36 +1,52 @@
...
+    // Set the time preference value once we load it from the server
+    useEffect(() => {
+        if (timePref) {
+            setTimeSelect(timePref);
+        }
+    }, [ timePref ]);
 
-    const handleRadioSelect = React.useCallback(() => {
+    const handleRadioSelect = React.useCallback((time) => {
+        setTimeSelect(time);
```
josejulio: What is this `time` value? Is this the default time? where does it come from?

src/components/Notifications/TimeConfig.tsx
```diff
@@ -45,26 +61,27 @@ export const TimeConfigComponent: React.FunctionComponent = () => {
...
-    }, []);
+    }, [ ]);
+
+    const handleButtonSave = React.useCallback(() => {
+        if (timeSelect) {
+            const mutate = saveTimePreference.mutate;
+            mutate({
+                body: timeSelect
+            }).then((response) => {
+                if (response.status === 200) {
```
josejulio: This is changing to 204 (No content) - I'll prepare a followup PR to sync after we merge https://github.com/RedHatInsights/notifications-backend/pull/2069

src/components/Notifications/TimeConfig.tsx
aferd: this was looking to save the default time, but looking again it should be body as the prop. If someone selects the default time, i was not sending that default time.

aferd: well, that is also not right... the body is returning a boolean...

aferd: Not needed as the value is being sent when "save" is hit.

src/components/Notifications/TimeConfig.tsx
```diff
@@ -45,26 +60,27 @@ export const TimeConfigComponent: React.FunctionComponent = () => {
...
```
josejulio: Lets change this to 204 - the other PR has been merged
```suggestion
                if (response.status === 204) {
```

josejulio (approved):
lgtm

PR #430, opened 2023-06-27
RHCLOUD-26700 | feature: move Event Driven Ansible to "stable"

MikelAlejoBR:
Since Red Hat Insights EDA v1.0.0 is available on the Automation Hub, we can consider this integration as "stable" now.

## Links

[[RHCLOUD-26700]](https://issues.redhat.com/browse/RHCLOUD-26700)

### DIFF
--- a/src/config/Config.ts
+++ b/src/config/Config.ts
@@ -101,6 +101,7 @@ const Config = {
         types: computeIntegrationConfig(integrationTypes),
         actions: {
             stable: sortedIntegrationList([
+                UserIntegrationType.ANSIBLE,
                 UserIntegrationType.GOOGLE_CHAT,
                 UserIntegrationType.TEAMS,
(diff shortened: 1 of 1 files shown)


src/config/Config.ts
```diff
@@ -118,7 +118,6 @@ const Config = {
                 UserIntegrationType.WEBHOOK
             ]),
             experimental: sortedIntegrationList([
-                UserIntegrationType.ANSIBLE,
```
josejulio: It's better to leave them here and in beta, they are not inherited.

MikelAlejoBR: Aaaah, I though we had to remove them from `experimental` and `beta` and only leave them in `stable`. All right, corrected!

josejulio (approved):

PR #443, opened 2023-08-04
Notifications Landing Page

radekkaluzik:
https://issues.redhat.com/browse/RHCLOUD-25994


src/assets/icons/notifications-product-icon.svg
karelhala, line 1: We should probably move these files to static assets repository.

karelhala (approved):
Looking good!

PR #444, opened 2023-08-08
Enable overhaul navigation via feature flags

karelhala:
### Desctiption

Change navigation based on active flag and react to it by selecting correct route.

I've also removed legacy `chrome.on` function which shouldn't be used anymore (once we migrate over to react router v6 it can actually break a lot of interactions)


Hyperkid123 (approved):

josejulio:
Probably missing the mock on the smoke test: https://github.com/RedHatInsights/notifications-frontend/blob/master/test/SmokeTest.tsx#L1 and one test is complaining that this dependency is not pinned: https://github.com/RedHatInsights/notifications-frontend/pull/444/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R17

PR #450, opened 2023-08-23
Add chrome on events reaction

karelhala:
### Description

When user clicks on navigation the UI does not respond to it, because we are missing reaction to chrome.on events.


Hyperkid123 (commented):
I assume it won't be easier to just use router v6 right?

karelhala:
@Hyperkid123 this is a hotfix and we have react router v6 issue planned for this/next sprint

Hyperkid123 (approved):

PR #451, opened 2023-08-24
RHCLOUD-22647 update event-log filters to show correct status

aferd:
Before - 
Screen Shot 2022-11-17 at 10 44 14 AM

After - 
Screen Shot 2023-08-24 at 1 54 07 PM

### DIFF
--- a/src/components/Notifications/EventLog/EventLogToolbar.tsx
+++ b/src/components/Notifications/EventLog/EventLogToolbar.tsx
@@ -1,5 +1,6 @@
 import { PaginationProps, PaginationVariant } from '@patternfly/react-core';
 import { CheckCircleIcon, ExclamationCircleIcon, ExclamationTriangleIcon, InProgressIcon } from '@patternfly/react-icons';
+import { global_warning_color_100 } from '@patternfly/react-tokens';
 import { PrimaryToolbar } from '@redhat-cloud-services/frontend-components';
 import { ConditionalFilterProps } from '@redhat-cloud-services/frontend-components/ConditionalFilter';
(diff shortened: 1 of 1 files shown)


src/components/Notifications/EventLog/EventLogToolbar.tsx
```diff
@@ -61,6 +61,22 @@ const actionStatusMetadata = [
...
     },
+    {
+        value: 'false',
+        chipValue: 'Sent',
+        label: <span><CheckCircleIcon color='green' /> Sent</span>
+    },
+    {
+        value: 'false',
+        chipValue: 'Warning',
+        label: <span><ExclamationTriangleIcon color='#F0AB00' /> Warning </span>
```
karelhala: Might be good if we could use CSS var `--pf-chart-donut--threshold--warning--Color` and custom class I am just not sure if it's applicable to icons.

karelhala (approved):
looking good! Just one nitpick about inline color.

src/components/Notifications/EventLog/EventLogToolbar.tsx
```diff
@@ -55,12 +57,32 @@ const notificationTypes: Record<NotificationType, { name: string }> = {
     }
 };
 
+const warningIconColorClassName = style({
```
aferd, line 60: like this? I found the global_color var for that color in PF.

karelhala: Ah, we are using `typestyle` for styling. Yeah, this is the correct way!

karelhala (approved):

PR #452, opened 2023-09-07
Tab config event

InsaneZein:
Part of [RHCLOUD-27097](https://issues.redhat.com/browse/RHCLOUD-27097)

configure_events

### DIFF
--- a/src/pages/Notifications/List/BundlePage.tsx
+++ b/src/pages/Notifications/List/BundlePage.tsx
@@ -6,20 +6,23 @@ import {
     localUrl
 } from '@redhat-cloud-services/insights-common-typescript';
 import { useFlag } from '@unleash/proxy-client-react';
-import { default as React } from 'react';
+import { default as React, useMemo } from 'react';
 
 import { useAppContext } from '../../../app/AppContext';
+import { AppSkeleton } from '../../../app/AppSkeleton';
 import { ButtonLink } from '../../../components/ButtonLink';
 import { TabComponent } from '../../../components/Notifications/TabComponent';
 import { TimeConfigComponent } from '../../../components/Notifications/TimeConfig';
 import { PageHeader } from '../../../components/PageHeader';
 import { Messages } from '../../../properties/Messages';
 import { linkTo } from '../../../Routes';
+import { useGetApplicationsLazy } from '../../../services/Notifications/GetApplications';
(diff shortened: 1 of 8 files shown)


karelhala:
@InsaneZein looks like the tests needs a bit of adjustment

InsaneZein:
I'm currently working on them

src/pages/Notifications/List/BundlePage.tsx
```diff
@@ -45,29 +49,85 @@ export const NotificationListBundlePage: React.FunctionComponent<NotificationLis
...
+                title={ `${Messages.pages.notifications.list.title} | ${props.bundle.displayName}` }
+                subtitle={ <span>This service allows you to configure which notifications different
+                    users within your organization will be entitled to receiving. To do this, create behavior groups and apply
+                    them to different events. Users will be able to opt-in or out of receiving authorized event notifications in their
+                <a href={ localUrl(`/user-preferences/notifications/${props.bundle.name}`,
+                    getInsights().chrome.isBeta()) }> User Preferences</a>.</span> }
+                action={ eventLogButton() }
+            />
+
+            <TabComponent configuration={ props.children } settings={ props.children }>
```
aferd, line 128: where did this section go? I don't see the time config in the mocks but curious if that was on purpose or not.

InsaneZein: I think this is supposed to go to a different page. I don't remember where though. We might need a story for that.

aferd:
looks good! just one picky thing would be to line up the bundle tabs with the tabs below, similar to this.. so its stacked. (I got this review when I worked those initial tabs, so just a guess they would want the same thing!)
Screen Shot 2023-09-21 at 4 12 12 PM

src/pages/Notifications/List/BundlePage.tsx
aferd: yeah, worth reaching out to UXD.

InsaneZein:
Good catch. I fixed it
image
****

src/pages/Notifications/List/BundlePage.tsx
InsaneZein: I agree. Will do.

aferd:
looks like you need to run a lint fix after that lgtm!

PR #465, opened 2023-10-02
Activate configure events button

karelhala:
### Description

As a user when I see overview page I should be able to click on configure events button and be brought to `/settings/nottifications/configure-events` page.


Hyperkid123 (commented):
@karelhala there is a lint error

PR #469, opened 2023-10-06
Show landing page for non org admins

karelhala:
### Description

If user somehow navigates to page which they don't have access to, let's show overview page for them.


aferd (changes requested):
looks like some routes need to be updated in the tests.

after that lgtm!

PR #470, opened 2023-10-10
Upgrade to react-router-dom@v6

karelhala:
### Description

Let's upgrade this application to use react router dom v6. This PR also fixes a couple of blank navigations within overview and when old navigation is used.

### JIRA
https://issues.redhat.com/browse/RHCLOUD-28490


Hyperkid123:
@karelhala the tests will need some attention.

karelhala:
@Hyperkid123 tests are passing again!

src/Routes.tsx
```diff
@@ -71,58 +67,22 @@ const routesOverhaul: Path[] = [
...
-        <Switch>
+        <DomRoutes>
             { pathRoutes.map(pathRoute => (
-                <InsightsRoute
-                    key={ pathRoute.path }
-                    component={ pathRoute.component }
-                    path={ pathRoute.path }
-                    exact={ true }
-                />
+                <Route key={ pathRoute.path } path={ pathRoute.path } element={ <CheckReadPermissions>
```
Hyperkid123, line 78: @karelhala  Just curious about the `pathRoute.path`. As routes are `exact` by default now, I am wondering if the paths have the `/*` suffix. I am too lazy to check the code and see for myself :smile:

src/Routes.tsx
```diff
@@ -71,58 +67,22 @@ const routesOverhaul: Path[] = [
...
+                    <pathRoute.component />
+                </CheckReadPermissions> } />
             ))}
-            {!notificationsOverhaul && <RedirectToDefaultBundle />}
-        </Switch>
+            {!notificationsOverhaul && <Route
+                path="*"
```
Hyperkid123, line 83: I am in favor of actually using 404 pages instead of unexpected redirects.

src/app/useApp.ts
```diff
@@ -39,13 +33,12 @@ export const useApp = (): Partial<AppContext> => {
     }, [ serverStatus.payload ]);
 
     useEffect(() => {
-        waitForInsights().then(insights => {
-            insights.chrome.auth.getUser().then(user => {
-                setOrgAdmin(user.identity.user.is_org_admin);
-                fetchRBAC(`${Config.notifications.subAppId},${Config.integrations.subAppId}`).then(setRbac);
-            });
+        chrome.auth.getUser().then(user => {
+            console.log(user, 'this is user!');
```
Hyperkid123: console log

src/Routes.tsx
karelhala: I am as well, this is just backwards compatible layer. Notice the `notificationsOverhaul` once the notif overhaul will be GA this will be removed.

src/Routes.tsx
karelhala: Good point, none of the routes have nested details. So it's fine to keep as is. I've checked it with both legacy routes and the new ones and everything looked to be correct.

fhlavac (approved):
If @Hyperkid123 is ok with the changes, LGTM!

Hyperkid123 (approved):

PR #486, opened 2023-11-02
wrap timeconfig in redux provider for use in user-pref

InsaneZein:


src/components/Notifications/ConnectedTimeConfig.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+
+import { AppEntryProps } from '../../AppEntry';
+import { getNotificationsRegistry } from '../../store/Store';
+import { TimeConfigComponent } from './TimeConfig';
+
+export const ConnectedTimeConfig: React.FunctionComponent<AppEntryProps> = (props) => {
+
+    const store = React.useMemo(() => {
```
karelhala: This will create a new store, you don't need to make a new one. You can consume store from props of this component passed from user-preferences app.

karelhala (commented):
You'll want to expose this component trough the `fec` config.

src/components/Notifications/ConnectedTimeConfig.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+import { Store } from 'redux';
+
+import { TimeConfigComponent } from './TimeConfig';
+
+export const ConnectedTimeConfig = (store: Store) => {
```
karelhala: It's a prop so you have to wrap it in object.
```suggestion
export const ConnectedTimeConfig = ({store}: {store: Store}) => {
```

karelhala (approved):

PR #490, opened 2023-11-09
Update lint to use common rhcloud eslint config

florkbr:
This will update the notifications frontend repo to use our common rhcloud eslint config.

I tried splitting the effort up into 2 commits - the first makes most of the dependency/config changes, and the second makes the actual fixes to the codebase.

For some rules - for example `no-unexpected-any` - I decided to ignore every instance in the codebase file by file versus globally. Why? This will discourage use in the future as the rule is still on globally for future PRs/new files.

I'd like to propose as developers touch files in their PRs in the future - if they see ignored rules they should attempt to fix them.


config/karma.conf.ts
```diff
@@ -1,117 +1,116 @@
+// eslint-disable-next-line rulesdir/disallow-fec-relative-imports
```
florkbr, line 1: @Hyperkid123 FYI I had to ignore this suggested change (at least until we upgrade the version of `frontend-components-config` in package.json) as the rule was trying to suggest `@redhat-cloud-services/frontend-components-config/config` however that does not exist (in this version the config function is exported at the root).

package.json
```diff
@@ -54,6 +54,7 @@
     "@babel/plugin-transform-runtime": "7.12.1",
     "@babel/preset-env": "7.12.1",
     "@babel/preset-react": "7.12.5",
+    "@redhat-cloud-services/eslint-config-redhat-cloud-services": "^1.3.0",
```
florkbr, line 57: We can bump this to ^2.0.0 once notifications-frontend migrates to PF5

package.json
```diff
@@ -54,6 +54,7 @@
     "@babel/plugin-transform-runtime": "7.12.1",
     "@babel/preset-env": "7.12.1",
     "@babel/preset-react": "7.12.5",
+    "@redhat-cloud-services/eslint-config-redhat-cloud-services": "1.3.0",
```
florkbr, line 57: We can bump this to ^2.0.0 once notifications-frontend migrates to PF5

Hyperkid123 (commented):
@florkbr looks, good. I'd like to wait for @radekkaluzik and for the current PR he is working now. There changes will cause many conflicts with his work. And it will be easier for you to rebase and fix any linting issues than the other way around.

Hyperkid123 (approved):

src/pages/Notifications/List/BehaviorGroupsSection.tsx
```diff
@@ -31,250 +35,298 @@ import { DeleteBehaviorGroupPage } from '../Form/DeleteBehaviorGroupPage';
...
-                            props.behaviorGroupContent.content.length > 0 ?
-                                <Badge className={ badgeClassName } isRead>{props.behaviorGroupContent.content.length}</Badge> :
-                                <BehaviorGroupAddButton
-                                    className={ emptyAddButtonClassName }
-                                    component='a'
-                                    onClick={ createGroup }
-                                    isDisabled={ !rbac.canWriteNotifications }
-                                />
-                        )}
+                      <SearchInput
```
aneelac22, line 276: @florkbr^^

PR #492, opened 2023-11-10
RHCLOUD-28959 Reporting Tab

aferd:


src/pages/Integrations/Create/schema.ts
```diff
@@ -18,6 +18,7 @@ export const schema = (category) => ({
         integrationsStep(category),
         {
           title: 'Enter details',
+          label: 'Enter the details for your integration.',
```
InsaneZein, line 33: Oh we should move the details step to a different file since there's going to be some conditionals.

src/pages/Integrations/Create/schema.ts
```diff
@@ -72,8 +83,11 @@ export const schema = (category) => ({
                     name: 'secret-token',
                     type: 'text',
                     label: 'Secret token',
-                    helperText:
-                      'The defined secret token is sent as a "X-Insight-Token" header on the request.',
+                    helperText: UserIntegrationType.SERVICE_NOW
```
aferd, line 86: this piece is WIP

src/pages/Integrations/Create/IntegrationsStep.tsx
```diff
@@ -14,12 +14,11 @@ export const integrationsStep = (category: string) => ({
...
       component: componentTypes.PLAIN_TEXT,
       name: 'plain-text',
-      label:
-        'Select a communications integration to add to your Red Hat account.',
+      label: `Select a ${category.toLowerCase()} integration to add to your Red Hat account.`,
     },
     {
       component: CARD_SELECT,
-      name: 'communication-type',
+      name: 'category-type',
```
InsaneZein:
```suggestion
      name: 'integration-type',
```

InsaneZein (approved):
one small suggestion. Looks good otherwise

src/pages/Integrations/Create/schema.ts
```diff
@@ -72,8 +83,12 @@ export const schema = (category) => ({
                     name: 'secret-token',
                     type: 'text',
                     label: 'Secret token',
-                    helperText:
-                      'The defined secret token is sent as a "X-Insight-Token" header on the request.',
+                    helperText: IntegrationType.SERVICE_NOW
```
aferd: I double checked and this piece isn't working. @InsaneZein any chance you could check it out? I feel like its pretty straight forward but its only showing the ServiceNow text.

InsaneZein: oops I overlooked this. This is always going to return true. We might need something like this https://www.data-driven-forms.org/schema/condition-set

PR #493, opened 2023-11-13
Add DopeBox to Integrations tabs

epwinchell:
Jira: https://issues.redhat.com/browse/RHCLOUD-28955

TODO:

- [ ] Show appropriate link content by tab

Screenshot 2023-11-13 at 4 11 16 PM


src/components/Integrations/DopeBox.tsx
```diff
@@ -0,0 +1,120 @@
...
+import '../../app/App.scss';
+
+const DopeBox = () => {
+  const [isExpanded, setIsExpanded] = React.useState(false);
+  const onToggle = (_event: React.MouseEvent, isExpanded: boolean) => {
+    setIsExpanded(isExpanded);
+  };
+  return (
+    <Card className="noti-c-card-getting-started pf-v5-u-mb-xl">
+      <ExpandableSection className="pf-u-w-100 pf-u-flex-direction-row" toggleText={isExpanded ? 'Making use of Integrations' : 'Making use of Integrations'} onToggle={onToggle} isExpanded={isExpanded}>
```
aferd: Since the toggle text doesn't change, it shouldn't need the condition.
```suggestion
      <ExpandableSection className="pf-u-w-100 pf-u-flex-direction-row" toggleText='Making use of Integrations' onToggle={onToggle} isExpanded={isExpanded}>
```

src/components/Integrations/DopeBox.tsx
```diff
@@ -0,0 +1,120 @@
...
+              </CardHeader>
+              <CardBody className="pf-v5-u-display-flex">
+                <TextContent className="pf-v5-u-display-flex pf-v5-u-flex-direction-column pf-v5-u-align-self-stretch">
+                  <Text component={TextVariants.p} className="pf-v5-u-mb-sm pf-v5-u-active-color-100">
+                    <Icon className="pf-v5-u-pl-sm pf-v5-u-pr-md"><CogIcon /></Icon>Configure Applications
+                  </Text>
+                  <Text component="p" className="pf-u-flex-grow-1">
+                    To prepare for integration with the Hybrid Cloud Console, you must configure incoming webhooks in your third-party applications.
+                  </Text>
+{/*               <TextList className="pf-u-ml-0">
```
aferd: one thought would be possibly moving all the text to /properties/messages just to not clog the page and make accessing them a bit easier. 

cc: @fhlavac @karelhala

epwinchell: The markup will also be a bit different between tabs (see screenshot)

karelhala: @aferd looks like there are links and some text styling along side with the text, what we could do is create components for each section and map those based on the active tab.

PR #497, opened 2023-11-16
add communications steps

InsaneZein:


src/pages/Integrations/Create/helpers.tsx
```diff
@@ -47,3 +48,16 @@ export const compileAllIntegrationComboOptions = (
       label: t.product_name,
     }));
 };
+
+export const nextDetailsStep = (values) => {
```
InsaneZein, line 52: There's probably a better way to conditionally show the fields, but we need to get this done today. I'll clean this up during tech-debt.

aferd (approved):

PR #499, opened 2023-11-20
Integrating dopebox based on category.

arivepr:
Final bit for: [Integrations Jira Card](https://issues.redhat.com/browse/RHCLOUD-28955).

Only shows additional text depending on category.
Screenshot 2023-11-20 at 4 02 16 AM

Screenshot 2023-11-20 at 4 02 30 AM

Screenshot 2023-11-20 at 4 02 43 AM

### DIFF
--- a/src/components/Integrations/DopeBox.tsx
+++ b/src/components/Integrations/DopeBox.tsx
@@ -18,9 +18,14 @@ import { CogIcon } from '@patternfly/react-icons';
 import { LockIcon } from '@patternfly/react-icons';
 import { BellIcon } from '@patternfly/react-icons';
 import { ArrowRightIcon } from '@patternfly/react-icons';
+import { IntegrationCategory } from '../../types/Integration';
 import '../../app/App.scss';
 
-const DopeBox = () => {
+interface DopeBoxProps {
+  category?: IntegrationCategory;
+}
+
+const DopeBox: React.FC<DopeBoxProps> = ({ category }) => {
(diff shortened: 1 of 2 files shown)


arivepr:
Pushed it through given this is more of a Hotfix than anything else. Any comments, or concerns, more than happy to respond through here or Slack.

fhlavac:
@arivepr just checked with the mockups and the text below "Configure notifications" seems to be incorrect for Reporting and Webhooks.

Also in the "Configure applications" sections, could you make the list bullet points blue?

src/components/Integrations/DopeBox.tsx
```diff
@@ -67,72 +72,75 @@ const DopeBox = () => {
...
-                        target="_blank"
-                      >
-                        Configure Microsoft Teams
-                      </Link>
-                    </TextListItem>
-                  </TextList>
-                  (Reporting tab)
-                  <TextList className="pf-u-ml-0">
-                    <TextListItem>
+                  {category === IntegrationCategory.COMMUNICATIONS && (
```
fhlavac, line 75: we could clean this up a bit

idea 1) put links and their labels to an array of objects and map it instead of manually copying the TextListItem
idea 2 - more optional) create a mapper that would return appropriate "Configure applications" links according to a category passed

arivepr:
As per our slack conversation, I'll put these on a ticket and treat as a separate task entirely.

PR #505, opened 2023-11-28
RHCLOUD-28952: Add empty integrations state

florkbr:
### As non org-admin
Screenshot 2023-11-27 at 11 15 21 PM

### As org-admin
Screenshot 2023-11-27 at 11 18 30 PM

### Responsiveness
Screenshot 2023-11-27 at 10 57 32 PM


src/pages/Integrations/Create/CreateWizard.tsx
```diff
@@ -34,7 +34,9 @@ export const CreateWizard: React.FunctionComponent<CreateWizardProps> = ({
   return isOpen ? (
     <FormRenderer
       schema={schema(category)}
-      FormTemplate={Pf4FormTemplate}
+      FormTemplate={(props) => (
+        <Pf4FormTemplate {...props} showFormControls={false} />
```
florkbr, line 38: This is a bug I noticed where opening the `Add Integration` wizard causes submit buttons to show in the UI behind the modal.

src/components/Integrations/EmptyState.tsx
```diff
@@ -0,0 +1,112 @@
...
+            aria-label="Card container"
+            minWidths={{ default: '100%', md: isOrgAdmin ? '30%' : '40%' }}
+            hasGutter
+            className="pf-v5-u-text-align-left"
+          >
+            {getEmptyIntegrationCard({
+              TitleIcon: HelpIcon,
+              title: 'Why integrate?',
+              body: 'Integrating third-party applications expands the scope of notifications beyond emails and messages, so that you can view and manage Hybrid Cloud Console events from your preferred platform dashboard.',
+              link: 'https://access.redhat.com/documentation/en-us/red_hat_hybrid_cloud_console/2023/html/integrating_the_red_hat_hybrid_cloud_console_with_third-party_applications',
```
florkbr: I've taken my best guess for where these links should navigate.

src/components/Integrations/EmptyState.tsx
```diff
@@ -0,0 +1,112 @@
...
+    <div className="pf-v5-l-flex pf-m-justify-content-center">
+      <EmptyState>
+        <EmptyStateIcon icon={CubesIcon} />
+        <Title headingLevel={TextVariants.h4} size={TitleSizes.lg}>
+          No integrations yet
+        </Title>
+        <EmptyStateBody>
+          <Gallery
+            aria-label="Card container"
+            minWidths={{ default: '100%', md: isOrgAdmin ? '30%' : '40%' }}
```
florkbr, line 86: This styling is not ideal - but works as expected. I've tried switching to flex, however that has it's own set of bugs (odd wrapping) that will take some time to figure out. FWIW `minWidths` adjusts the number of columns rendered and `maxWidth` in the card above controls how large the card can grow.

fhlavac (approved):
Looks good!

florkbr:
There is still a bug where the integrations table still momentarily flashes before the empty state. I suspect that is an issue with our `isLoading` logic. I will follow up on that separately.

InsaneZein (approved):

src/components/Integrations/EmptyState.tsx
```diff
@@ -0,0 +1,119 @@
...
+  InfrastructureIcon,
+} from '@patternfly/react-icons';
+import React from 'react';
+
+import { Messages } from '../../properties/Messages';
+import { Link } from 'react-router-dom';
+import { useAppContext } from '../../app/AppContext';
+
+const getEmptyIntegrationCard = (props) => {
+  const { TitleIcon, title, body, link, isOrgAdmin } = props;
```
karelhala: I'd spread the props directly when using the `props`. Also the name is not ideal, it does not indicate it's going to produce React component.
```suggestion
const EmptyIntegrationCard = ({ TitleIcon, title, body, link, isOrgAdmin }) => {
```

src/components/Integrations/EmptyState.tsx
```diff
@@ -0,0 +1,119 @@
...
+            className="pf-v5-u-text-align-left"
+          >
+            {getEmptyIntegrationCard({
+              isOrgAdmin,
+              TitleIcon: HelpIcon,
+              title: 'Why integrate?',
+              body: 'Integrating third-party applications expands the scope of notifications beyond emails and messages, so that you can view and manage Hybrid Cloud Console events from your preferred platform dashboard.',
+              link: 'https://access.redhat.com/documentation/en-us/red_hat_hybrid_cloud_console/2023/html/integrating_the_red_hat_hybrid_cloud_console_with_third-party_applications',
+            })}
+            {getEmptyIntegrationCard({
```
karelhala: It would be nicer if we'd change the `getEmptyIntegrationCard` to a regular component.

src/pages/Integrations/List/List.tsx
```diff
@@ -225,6 +226,50 @@ const IntegrationsList: React.FunctionComponent<IntegrationListProps> = ({
...
+  }
+
+  if (deleteModalState.data) {
+    integrationsModals.push(
+      <IntegrationDeleteModalPage
+        onClose={closeDeleteModal}
+        integration={deleteModalState.data}
+      />
+    );
+  }
```
karelhala: I am not a big fan of this statement, how about we'll create a conditional array? I know it's a bit more expensive than `if` and `push` but we could wrap this into a helper function and that way it will be a bit more readable.

Looking at the code further down the line I would keep the code as was previously with `if`s in JSX and wrap the rest with another condition. I know it might sound a bit cumbersome and clunky, but if we'd were to keep the array we'd be wondering why it's in both places. I overlooked it as well on initial look.

```JSX
const integrationsModals = [
  ...(modalIsOpenState.isOpen && !wizardEnabled) ? [<CreatePage
        isEdit={modalIsOpenState.isEdit}
        initialIntegration={modalIsOpenState.template || {}}
        onClose={closeFormModal}
      />] : [],
   ...wizardEnabled ? [<CreateWizard
        isOpen={modalIsOpenState.isOpen}
        closeModal={() => modalIsOpenActions.reset()}
        category={category}
      />] : [],
   ....deleteModalState.data ? [<IntegrationDeleteModalPage
        onClose={closeDeleteModal}
        integration={deleteModalState.data}
      />] : []
];
```

karelhala (commented):
Just two minor adjustments, but overall looks good!

src/components/Integrations/EmptyState.tsx
florkbr: @karelhala done!

src/components/Integrations/EmptyState.tsx
florkbr: @karelhala good catch. Done!

src/pages/Integrations/List/List.tsx
florkbr: @karelhala I'm not a big fan of conditional rendering in the JSX itself - especially as the JSX grows - it becomes much harder to debug what is rendered and why. It gets even worse as we add to the conditionals, and can result in unexpected things rendering as the conditions may be linked together. Whereas returning fast/immediately seems much clearer to me (paying close attention to avoid DOM duplication). 

IE (pseudo code, but like I had it):
```tsx
if (empty) {
  return <EmptyState .../>
}

return <IntegrationsUI .../>
```
This seems cleaner to me - and avoids the risk of accidentally rendering the IntegrationsUI and the EmptyState at the same time.

versus:
```tsx
return (
  <>
    {empty && (<EmptyState ... />)}
    {!empty && (<IntegrationsUI .../>)}
  </>
);
```

However I see we are doing that already in the codebase - and I'm not yet willing to lie down on the tracks for it. So for now I've updated to as you requested I think. 

Regarding your point of having the `integrationsModals` in two places being confusing - that is not strictly required. I did it that way to keep the PR size small as refactoring the modal setup logic and abstracting it outside of `List.tsx` is another effort in and of itself. I do think we should consider this blog post around how our components should be implemented: https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0 (sorry if I'm a broken record).

**TL;DR** - updated as you requested for now; we should revisit this style as a broader discussion.

karelhala (approved):

src/pages/Integrations/List/List.tsx
karelhala: Sure thing! I completely understand you, let's bring this up with the wider team and discuss what's easier for us to read and mainaint.

PR #506, opened 2023-11-28
Integrations  testing functionality

arivepr:
Addressing [the following Jira Card](https://issues.redhat.com/browse/RHCLOUD-25193)

Adding test functionality to integrations appearing in the integrations table, using internal pop-up notifications.

Screenshot 2023-11-28 at 11 54 50 AM

Screenshot 2023-11-28 at 11 55 32 AM

D815251C-D1A6-44D5-84EA-6BD6AA58DF14

As per previous discussion, leaving the Feature flag commented out for now given it forces

### DIFF
--- a/src/components/Integrations/Table.tsx
+++ b/src/components/Integrations/Table.tsx
@@ -351,8 +351,7 @@ export const IntegrationsTable: React.FunctionComponent<IntegrationsTableProps> 
     const actionsResolverCallback: IActionsResolver = React.useCallback(
       (rowData) => {
         const actionResolver = props.actionResolver;
-        console.log('Testing our rowData', rowData);
-        console.log('Testing our actionResolver', actionResolver);
+
         if (rowData.parent === undefined && rowData && props.integrations) {
           const integrationIndex = props.integrations.findIndex(
             (i) => i.id === rowData.id
--- a/src/components/Integrations/Table.tsx
+++ b/src/components/Integrations/Table.tsx
@@ -351,6 +351,8 @@ export const IntegrationsTable: React.FunctionComponent<IntegrationsTableProps> 
     const actionsResolverCallback: IActionsResolver = React.useCallback(
       (rowData) => {
         const actionResolver = props.actionResolver;
+        console.log('Testing our rowData', rowData);
+        console.log('Testing our actionResolver', actionResolver);
         if (rowData.parent === undefined && rowData && props.integrations) {
           const integrationIndex = props.integrations.findIndex(
             (i) => i.id === rowData.id
--- a/src/components/Integrations/Table/IntegrationTest.tsx
+++ b/src/components/Integrations/Table/IntegrationTest.tsx
@@ -1,14 +1,12 @@
 import React, { useState } from 'react';
 import axios from 'axios';
 import { Button, Modal, TextInput } from '@patternfly/react-core';
-import {
-  addSuccessNotification,
-  addWarningNotification,
-} from '@redhat-cloud-services/insights-common-typescript';
+import { useNotification } from '../../../utils/AlertUtils';
 
 const IntegrationTestModal = ({ integrationUUID }) => {
   const [isModalOpen, setIsModalOpen] = useState(true);
   const [inputValue, setInputValue] = useState('');
+  const { addSuccessNotification, addWarningNotification } = useNotification();
 
   const placeholderText =
     'Congratulations! The integration you created on https://console.redhat.com was successfully tested!';
@@ -27,11 +25,12 @@ const IntegrationTestModal = ({ integrationUUID }) => {
 
     try {
       const response = await axios.post(endpointURL, body);
-      // response?.status === 204
-      //   ? addSuccessNotification('Integration Test', '')
-      //   : addWarningNotification('Failed Test', '');
-
-      addSuccessNotification('Integration Test', notificationMessage);
+      response?.status === 204
+        ? addSuccessNotification('Integration Test', notificationMessage)
+        : addWarningNotification(
+            'Failed Test',
+            `Error through server response: ${response.status}`
+          );
 
       console.log('Succesful request. Response data:', response);
     } catch (error) {
--- a/src/components/Integrations/Table/IntegrationTest.tsx
+++ b/src/components/Integrations/Table/IntegrationTest.tsx
@@ -1,30 +1,75 @@
 import React, { useState } from 'react';
-import { Button, Modal } from '@patternfly/react-core';
+import axios from 'axios';
+import { Button, Modal, TextInput } from '@patternfly/react-core';
+import {
+  addSuccessNotification,
+  addWarningNotification,
+} from '@redhat-cloud-services/insights-common-typescript';
 
-const IntegrationTestModal = () => {
-    const [ isModalOpen, setIsModalOpen ] = useState(false);
+const IntegrationTestModal = ({ integrationUUID }) => {
+  const [isModalOpen, setIsModalOpen] = useState(true);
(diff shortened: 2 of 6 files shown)


arivepr:
Just a quick update: Currently exploring wrapping the component with the notificationRegistry. Had not realized we're using registry on this repo, trying to see if by explicitly wrapping this component specifically, the Pop ups will then start rendering properly.

Re-update: Did not fix it :D

src/components/Integrations/Table.tsx
```diff
@@ -351,6 +351,7 @@ export const IntegrationsTable: React.FunctionComponent<IntegrationsTableProps>
     const actionsResolverCallback: IActionsResolver = React.useCallback(
       (rowData) => {
         const actionResolver = props.actionResolver;
+
```
florkbr, line 354: @arivepr nit - while I know this _looks_ more sensible than it did with the extra new line - I usually try to avoid checking in these changes with other code as it could cause merge conflicts with someone else working on this file (especially if there are no other changes to the file in your PR).

src/pages/Integrations/List/List.tsx
```diff
@@ -191,10 +202,12 @@ const IntegrationsList: React.FunctionComponent<IntegrationListProps> = ({
   const actionResolver = useActionResolver({
     canWrite: canWriteIntegrationsEndpoints,
     onEdit,
+    onTest,
     onDelete,
     onEnable: integrationRows.onEnable,
   });
 
+  // eslint-disable-next-line
```
florkbr, line 210: @arivepr is this necessary? It looks like this line was passing linter before?

src/components/Integrations/Table/IntegrationTestProvider.tsx
florkbr, line 1: nit and not blocking - I wish we used another word internally to refer to this feature other than `Test` as upon initial glance I thought this was a test file and not production code/functionality.

src/components/Integrations/Table/IntegrationTest.tsx
```diff
@@ -0,0 +1,71 @@
...
+import axios from 'axios';
+import { Button, Modal, TextInput } from '@patternfly/react-core';
+import { useNotification } from '../../../utils/AlertUtils';
+
+const IntegrationTestModal = ({ integrationUUID, isModalOpen, onClose }) => {
+  const [inputValue, setInputValue] = useState('');
+  const { addSuccessNotification, addWarningNotification } = useNotification();
+
+  const placeholderText =
+    'Congratulations! The integration you created on https://console.redhat.com was successfully tested!';
```
florkbr, line 11: @arivepr hardcoding `https://console.redhat.com` may not be ideal as this will also appear on `https://console.stage.redhat.com`?

src/components/Integrations/Table/IntegrationTest.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const failedTestText =
+    'Your integration test has unfortunately failed, please verify your integration information.';
+
+  const handleModalCancel = () => {
```
InsaneZein: Why wrap a function in a function? It doesn't seem necessary.

InsaneZein (approved):
just one lil comment. Looks good to me otherwise.

src/components/Integrations/Table/IntegrationTest.tsx
arivepr: Ah nice catch!  Pushed logic to the parent, never got rid of the original func lol

I'll clean it up in a sec.

src/components/Integrations/Table/IntegrationTest.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const IntegrationTestModal = ({ integrationUUID, isModalOpen, onClose }) => {
+  const [inputValue, setInputValue] = useState('');
+  const { addSuccessNotification, addWarningNotification } = useNotification();
+
+  const placeholderText =
+    'Congratulations! The integration you created on https://console.redhat.com was successfully tested!';
+
+  const failedTestText =
+    'Your integration test has unfortunately failed, please verify your integration information.';
```
florkbr, line 14: @arivepr where did we get these strings from?

src/components/Integrations/Table/IntegrationTest.tsx
florkbr: Also this seems a bit odd as placeholder text to me. It almost makes it feel like the integration is tested _before_ you submit the request. Did this come from UX?

src/components/Integrations/Table/IntegrationTest.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const handleNotificationTest = async (notificationMessage: string) => {
+    const endpointURL = `/api/integrations/v1/endpoints/${integrationUUID}/test`;
```
florkbr, line 17: Not unique to this PR - but we should try to move to a common route builder in case the root API version or path ever changes.

src/components/Integrations/Table/IntegrationTest.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    try {
+      const response = await axios.post(endpointURL, body);
+      response?.status === 204
+        ? addSuccessNotification('Integration Test', notificationMessage)
+        : addWarningNotification(
+            'Failed Test',
+            `Error through server response: ${response.status}`
+          );
+      onClose();
```
florkbr, line 29: @arivepr since `onClose();` is executed in both the `try` and `catch`, should we move it to `finally` instead and dedupe? Also I'd imagine this conditional logic is hit when the server returns 400 or 500 level errors. Is showing a notification with just the `response.status` actually enough information?

src/pages/Integrations/List/List.tsx
```diff
@@ -278,7 +295,14 @@ const IntegrationsList: React.FunctionComponent<IntegrationListProps> = ({
...
       )}
-      {wizardEnabled && category && (
+      {modalIsOpenState.isTest && (
+        <IntegrationTestProvider
+          integrationUUID={selectedIntegrationID}
+          onClose={() => setIsTestModalOpen(false)}
+          isModalOpen={isTestModalOpen}
+        />
+      )}
+      {wizardEnabled && category && !modalIsOpenState.isTest && (
```
florkbr, line 305: I'm not a huge fan of making these conditional rendering blocks even more complex than they already are - they do not make it abundantly clear what will render and why (making it harder and harder to diagnose if there is an issue). Also what happens if the delete or create modal are already open in this situation?

florkbr (commented):
This was merged before I could finish my review. Posting my comments anyway. Seeing styling issues on both stage and locally with the modal (too wide): 
Screenshot 2023-12-11 at 4 47 44 PM

It also seems the menu the "test" action appears in can be covered by some of our overlays in the far bottom right corner of the UI (when the "making use of integrations" is open). It's not unique to this PR - though we should investigate a better solution:
Screenshot 2023-12-11 at 4 47 16 PM

src/components/Integrations/Table/IntegrationTest.tsx
```diff
@@ -0,0 +1,67 @@
...
+      console.error('\nError sending test notification:', error);
+
+      const responseString = `${failedTestText} ${error}`;
+      addWarningNotification('Failed Test', responseString);
+      onClose();
+    }
+  };
+
+  return (
+    <Modal
```
karelhala, line 40: I think the modal should be 50 % or `medium` or something like that. It feels a bit odd to span across entire screen for just one input. PF has option to change the size, so we should utilize that.

PR #513, opened 2023-12-11
RHCLOUD-29459 add checkbox select and badge count

aferd:
[RHCLOUD-29459](https://issues.redhat.com/browse/RHCLOUD-29459)

### DIFF
--- a/src/components/Notifications/Table/BehaviorGroupCell.tsx
+++ b/src/components/Notifications/Table/BehaviorGroupCell.tsx
@@ -7,7 +7,6 @@ import {
   MenuItem,
   OptionsMenu,
   OptionsMenuItem,
-  OptionsMenuItemGroup,
   OptionsMenuToggle,
   Tooltip,
 } from '@patternfly/react-core';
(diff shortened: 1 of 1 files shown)


aferd:
Screenshot 2023-12-11 at 4 22 14 PM

src/components/Notifications/Table/BehaviorGroupCell.tsx
```diff
@@ -74,11 +79,19 @@ export const BehaviorGroupCell: React.FunctionComponent<BehaviorGroupCellProps>
...
+              found = props.behaviorGroupContent.content.find(
+                // eslint-disable-next-line testing-library/await-async-queries
+                findById(dataset.behaviorGroupId)
+              );
+            } else if (behaviorGroupId) {
+              found = props.behaviorGroupContent.content.find(
+                // eslint-disable-next-line testing-library/await-async-queries
+                findById(behaviorGroupId)
+              );
+            }
```
fhlavac: This could be simplified

```suggestion
          if (dataset.behaviorGroupId || behaviorGroupId) {
             const found = props.behaviorGroupContent.content.find(
                // eslint-disable-next-line testing-library/await-async-queries
                findById(dataset.behaviorGroupId || behaviorGroupId)
             );
```

fhlavac:
Looks good! 🎉 

Just the font size looks different from the mockups. We may need to confirm with the UX
image

Assume the "Mute event" and create button going to be added in a separate PR, right?

aferd:
> Looks good! 🎉
> 
> Just the font size looks different from the mockups. We may need to confirm with the UX image
> 
> Assume the "Mute event" and create button going to be added in a separate PR, right?

I will double check on the font here.

The mute event action and button are no longer needed since behavior groups are going away, so it will not be done.

fhlavac (approved):
@aferd if the font size is confirmed by UX, then LGTM

aferd:
confirmed by Bekah that it is ok as it is coming from PF.

PR #527, opened 2024-01-29
RHCLOUD-21407 A different filter group appears in Event Log when filtering by services

aferd:


src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -263,17 +263,11 @@ export const usePrimaryToolbarFilterConfigWrapper = (
...
-          chipValues &&
+
+        chipValues &&
           bundle.children?.some(
             (application) => !chipValues.includes(application.name)
-          )
-        ) {
-          chipValues.forEach((chipValue) => {
-            const applicationQueryParam = `${bundle.name}.${chipValue}`;
-            currApplicationFilters.push(applicationQueryParam);
```
fhlavac, line 274: I may be missing something, but as you removed the forEach, only two conditions producing unused boolean seem to be remaining there.

aferd: the customFilter is already iterating the chipValues above, removing this stops from re-iterating and adding this extra chip thats not needed, this is how I am interpreting it at least..

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
```diff
@@ -263,17 +263,11 @@ export const usePrimaryToolbarFilterConfigWrapper = (
         // Only add applications to Query Params under 2 conditions
         //     1. Bundle has children
         //     2. Every application under the Bundle is not selected
-        if (
-          chipValues &&
+
+        chipValues &&
           bundle.children?.some(
             (application) => !chipValues.includes(application.name)
```
fhlavac, line 269: of course, but we can definitely remove this as well, since it has no effect

src/components/Notifications/EventLog/usePrimaryToolbarFilterConfigWrapper.tsx
fhlavac: of course, but 
```
chipValues &&
          bundle.children?.some(
            (application) => !chipValues.includes(application.name)
          );
```
that remained in the code has no effect after your changes, it just return boolean value which is unused

fhlavac: also, after removing this line
`currApplicationFilters.push(applicationQueryParam);`
filter values for services are not reflected in the URL anymore. We probably want to keep this functionality

aferd: so digging into this a bit more.. the above code all becomes irrelevant - `applicationProducer` no longer produces anything, just an empty array. I am working on just updating the params according to the filter values for services.

aferd:
closing this ticket and following it up with another PR to fix the url params and refactoring here [RHCLOUD-31067](https://issues.redhat.com/browse/RHCLOUD-31067).

fhlavac:
@aferd sure, just since the applicationProducer's code is irrelevant, can we remove it? I just wouldn't like to merge dead code.

aferd:
> @aferd sure, just since the applicationProducer's code is irrelevant, can we remove it? I just wouldn't like to merge dead code.

of course! i will get rid of it and push that change up now.

fhlavac (approved):

PR #528, opened 2024-01-29
pass only connection string to table column

CodyWMitchell:
[RHCLOUD-27969](https://issues.redhat.com/browse/RHCLOUD-27969)
# Fixes for Integrations table
- Fix incorrect 'column-3' table heading on small screen
- Fix help popover not displaying correctly

### DIFF
--- a/src/components/Integrations/Table.tsx
+++ b/src/components/Integrations/Table.tsx
@@ -1,5 +1,11 @@
-import { EmptyStateVariant, Spinner, Switch } from '@patternfly/react-core';
-import { CubesIcon, HelpIcon } from '@patternfly/react-icons';
+import {
+  EmptyStateVariant,
+  Spinner,
+  Switch,
+  Text,
+  TextContent,
+} from '@patternfly/react-core';
+import { CubesIcon } from '@patternfly/react-icons';
 import { css } from '@patternfly/react-styles';
 import styles from '@patternfly/react-styles/css/components/Table/table';
 import {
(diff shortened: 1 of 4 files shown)


CodyWMitchell:
This PR addresses a bug stemming from the mixed-up use of PF4's composable table style and a different table style, which will be deprecated in PF5. (passing in the popover as a component or setting it's prop) Here I'm resolving this by correctly following the 'to-be-deprecated' table style. 

Given the composable table's resemblance to PF5's table implementation, we may want to consider either creating a ticket to remove the deprecated style now, or instead handle it as part of the PF5 upgrade.

CodyWMitchell:
Though this form of table is deprecated in PF5 and should not be used moving forward, it will not be removed any time soon. This will likely be replaced by the data-view component once it's ready, so the fix should be fine as-is for now.

src/components/Integrations/Table.tsx
```diff
@@ -202,19 +203,22 @@ const columns: Array<ICell> = [
...
+    title: Messages.components.integrations.table.columns.lastConnectionAttempt,
+    transforms: [
+      info({
+        popover: <LastConnectionHelpTable />,
+        popoverProps: {
+          hasAutoWidth: true,
+          headerContent: (
+            <TextContent>
+              <Text component="h6">
+                {'Last connection attempt status meanings'}
```
karelhala: No need to wrap in string expression
```suggestion
                Last connection attempt status meanings
```

src/components/TableHelpPopover/TableHelp.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <TableComposable variant={TableVariant.compact} borders={false}>
+      <Thead>
+        <Tr className={removeBorderBottomClass}>
+          <Th>Status</Th>
+          <Th>Meaning</Th>
+        </Tr>
+      </Thead>
+      <Tbody>
+        {props.tableBody.map((tb, index) => (
```
karelhala: It's better to directly break the element so we don't have to access it via indexes.
```suggestion
        {props.tableBody.map(([status, meaning], index) => (
```

src/components/TableHelpPopover/TableHelp.tsx
```diff
@@ -0,0 +1,41 @@
...
+          <Tr key={index}>
+            <Td>{tb[0]}</Td>
+            <Td>{tb[1]}</Td>
```
karelhala:
```suggestion
            <Td>{status}</Td>
            <Td>{meaning}</Td>
```

karelhala (dismissed):
Looking good! Just a few smaller nitpicks

karelhala (approved):

PR #593, opened 2024-08-14
RHCLOUD-34529 fix behavior group redirect to correct tab

aferd:
### Description

Selected bundle under 'product family' doesn't redirect to correct tab via 'View behavior group'

[RHCLOUD-34529](https://issues.redhat.com/browse/RHCLOUD-34529)

---

#### After:

---

### Checklist ☑️

##
##

### DIFF
--- a/src/pages/Integrations/Create/CustomComponents/CreatedStep.tsx
+++ b/src/pages/Integrations/Create/CustomComponents/CreatedStep.tsx
@@ -62,6 +62,9 @@ export const CreatedStep: React.FunctionComponent<ProgressProps> = (props) => {
             <Button
               variant="link"
               component="a"
+              href={`/${getBundle()}/notifications/configure-events?bundle=${
+                props.data.bundle_name
+              }&tab=behaviorGroups`}
               onClick={() => {
                 navigate(
                   `/${getBundle()}/notifications/configure-events?bundle=${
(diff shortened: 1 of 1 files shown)


src/pages/Integrations/Create/CustomComponents/CreatedStep.tsx
```diff
@@ -62,19 +62,10 @@ export const CreatedStep: React.FunctionComponent<ProgressProps> = (props) => {
...
                 props.data.bundle_name
               }&tab=behaviorGroups`}
               size="lg"
-              onClick={(e) => {
-                props.onClose();
-                e.preventDefault();
-                navigate(
-                  `/${getBundle()}/notifications/configure-events?tab=behaviorGroups`
-                );
-              }}
```
karelhala: We might want to keep the navigate active. So we don't perform full page refresh.

aferd: Since this appears after its created, I thought we would want a refreshed page to be sure we have the updated list of behavior groups?

karelhala: No need to trigger full page refresh, the page should pull the data when user lands on it via `navigate`

fhlavac (commented):
I would just keep also the href - the button will probably no longer support opening in a new tab via middle-click and it's maybe a bit cleaner approach.

fhlavac (approved):

PR #661, opened 2025-01-10
[RHCLOUD-33811] Move notification drawer to its own module

BlakeHolifield:
### Description

Creates a new component for the NotificationDrawer. The functionality was previously in the insights-chrome
repo and will now live here and be called as a `ScalprumComponent`

[RHCLOUD-33811](https://issues.redhat.com/browse/RHCLOUD-33811)

---

### Checklist ☑️

##
##


fec.config.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
   hotReload: process.env.HOT === 'true',
   routes: {
     ...(process.env.CONFIG_PORT && {
-      '/api/chrome-service/v1/static': {
+      '/api/chrome-service/v1/': {
```
Hyperkid123: Unless you want to run the entire backend for configs, you should keep the /static in the pathname

package.json
```diff
@@ -56,12 +60,15 @@
...
     "@babel/preset-react": "7.12.5",
+    "@cypress/code-coverage": "^3.12.9",
     "@redhat-cloud-services/eslint-config-redhat-cloud-services": "1.3.0",
     "@redhat-cloud-services/frontend-components-config": "^6.2.9",
     "@redhat-cloud-services/frontend-components-testing": "0.1.1",
     "@redhat-cloud-services/tsc-transform-imports": "^1.0.8",
     "@redhat-cloud-services/integrations-client": "^3.2.0",
     "@redhat-cloud-services/notifications-client": "^2.5.3",
+    "@scalprum/react-core": "^0.9.3",
+    "@simonsmith/cypress-image-snapshot": "^8.1.2",
```
Hyperkid123: I don't think you need the visual regressions testing packages and cypress config for your tests.

package.json
```diff
@@ -79,7 +86,8 @@
     "babel-plugin-lodash": "3.3.4",
     "babel-plugin-transform-imports": "2.0.0",
     "cross-fetch": "3.1.5",
-    "cypress": "^12.0.0",
+    "cypress": "^13.9.0",
+    "cypress-localstorage-commands": "^2.2.5",
```
Hyperkid123: Don't think this is needed as well. We are not doing any actual e2e testing.

package.json
```diff
@@ -56,12 +60,15 @@
     "@babel/plugin-transform-runtime": "7.12.1",
     "@babel/preset-env": "7.12.1",
     "@babel/preset-react": "7.12.5",
+    "@cypress/code-coverage": "^3.12.9",
```
Hyperkid123: This is not setup so I'd not install it.

package.json
```diff
@@ -119,6 +127,10 @@
   "scripts": {
     "build": "fec build",
     "static": "fec static",
+    "cypress": "cypress",
+    "cypress:open": "cypress open",
+    "cypress:run:e2e": "cypress run --browser electron",
```
Hyperkid123: No e2e tests exist

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+    useState(false);
+  const updateNotificationRead = useSetAtom(updateNotificationReadAtom);
+  const updateAllNotificationsSelected = useSetAtom(
+    updateNotificationsSelectedAtom
+  );
+  const [filterConfig, setFilterConfig] = useState<FilterConfigItem[]>([]);
+  const eventType: ChromeWsEventTypes =
+    'com.redhat.console.notifications.drawer';
+  const handleWsEvent = (event: ChromeWsPayload<NotificationsPayload>) => {
+    console.log('new event');
```
Hyperkid123: console.log

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+  const [hasNotificationsPermissions, setHasNotificationsPermissions] =
+    useState(false);
+  const updateNotificationRead = useSetAtom(updateNotificationReadAtom);
+  const updateAllNotificationsSelected = useSetAtom(
+    updateNotificationsSelectedAtom
+  );
+  const [filterConfig, setFilterConfig] = useState<FilterConfigItem[]>([]);
+  const eventType: ChromeWsEventTypes =
+    'com.redhat.console.notifications.drawer';
+  const handleWsEvent = (event: ChromeWsPayload<NotificationsPayload>) => {
```
Hyperkid123: The event payload should have the `NotificationData` as the generic type. You won't need the zas unknown as NotificationDataz in the `addNotification` call.

```
  const handleWsEvent = (event: ChromeWsPayload<NotificationData>) => {
```

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+    console.log('new event');
+    addNotification(event.data as unknown as NotificationData);
+  };
+
+  useEffect(() => {
+    let mounted = true;
+    const unregister = addWsEventListener(eventType, handleWsEvent);
+    const getNotifications = async () => {
```
Hyperkid123: This function should be fined outside of the useEffect

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+              startDate: getSevenDaysAgo(),
+            },
+          }
+        );
+        populateNotifications(data?.data || []);
+      } catch (error) {
+        console.error('Unable to get Notifications ', error);
+      }
+    };
+    const fetchPermissions = async () => {
```
Hyperkid123: ^^

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+            [
+              'notifications:*:*',
+              'notifications:notifications:read',
+              'notifications:notifications:write',
+            ].includes((typeof item === 'string' && item) || item?.permission)
+          )
+        );
+      }
+    };
+    const fetchFilterConfig = async () => {
```
Hyperkid123: ^^

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+            }))
+          );
+        }
+      } catch (error) {
+        console.error('Failed to fetch filter configuration:', error);
+      }
+    };
+    getNotifications();
+    fetchPermissions();
+    fetchFilterConfig();
```
Hyperkid123: These three should be wrapped in some "init" function and also handle the loading state of the notifications. Which we need if we want to fetch the data in this module.

Another issues is that right now we do not know if there are any "unread" notifications until this data is available. We fetch it only when the JS module is loaded. And that is only after someone opens the drawer.

The data fetching should probably be done in Chrome and passed down as a prop to the remote module. So we know if some new notification has arrived.

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+      mounted = false;
+      unregister();
+    };
+  }, []);
+
+  const filteredNotifications = useMemo(
+    () =>
+      (activeFilters || []).reduce(
+        (acc: NotificationData[], chosenFilter: string) => [
+          ...acc,
```
Hyperkid123: This will be constantly creating new arrays. The new instance of array is already defined as the second argument of reduce.
 The callback can use simple push
 ```js
(acc, chosenFilter ) => {
  acc.push(...notifications.filter //...)
  return acc
} 
```

Also having another nested loop in the reduce can be improved. You can prepare a simple map/object of the values with the required keys and then just access the values by it. No need for nested loops here

File: src/components/NotificationsDrawer/DrawerPanel.tsx [hunk 13]
```diff
@@ -0,0 +1,515 @@
...
+            </MenuToggle>
+          )}
+          isOpen={isFilterDropdownOpen}
+          onOpenChange={setIsFilterDropdownOpen}
+          popperProps={{
+            position: PopoverPosition.right,
+          }}
+          id="notifications-filter-dropdown"
+        >
+          {filterDropdownItems()}
```
Hyperkid123: This is not ideal. Can we create components from the render functions?

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+          popperProps={{
+            position: PopoverPosition.right,
+          }}
+          id="notifications-actions-dropdown"
+        >
+          <DropdownList>{dropdownItems}</DropdownList>
+        </Dropdown>
+      </NotificationDrawerHeader>
+      <NotificationDrawerBody>
+        <NotificationDrawerList>{renderNotifications()}</NotificationDrawerList>
```
Hyperkid123: Component would be better.

src/state/atoms/notificationDrawerAtom.ts
```diff
@@ -0,0 +1,81 @@
...
+  source: string;
+  // cloud events sub protocol metadata
+  datacontenttype: string;
+  specversion: string;
+  // a type field used to identify message purpose
+  type: string;
+  time: string;
+};
+
+export const notificationDrawerExpandedAtom = atom(false);
```
Hyperkid123: This should be handled in Chrome not here.

src/state/atoms/notificationDrawerAtom.ts
```diff
@@ -0,0 +1,81 @@
...
+export const notificationDrawerDataAtom = atom<NotificationData[]>([]);
+export const notificationDrawerCountAtom = atom(0);
```
Hyperkid123: I think the initial data should be passed from chrome as well as props. The notifications module should only consume them.

Hyperkid123 (commented):
Some of the logic should stay in Chrome OR we need to leverage some other techniques (federated modules 😉 ) so we can reach feature parity with the current approach.

package.json
```diff
@@ -28,6 +29,7 @@
     "formik": "2.2.9",
     "immer": "9.0.6",
     "in-browser-download": "2.0.0",
+    "jotai": "^2.10.1",
```
karelhala: Notifications application is using Redux. I don't think it's reasonable to add Jotai while adding a the drawer to it.

config/webpack.cy.config.js
karelhala, line 1: This PR should not add cypress config. We have a seperate ticket for that work.

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
+import React, { useEffect, useMemo, useState } from 'react';
+import { PopoverPosition } from '@patternfly/react-core/dist/dynamic/components/Popover';
+import { Badge } from '@patternfly/react-core/dist/dynamic/components/Badge';
+import { useAtom, useAtomValue, useSetAtom } from 'jotai';
```
karelhala: We are using redux in notifications application, so it might be better to change the state logic to it so we don't have 2 different state management systems.

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+import { Link, useNavigate } from 'react-router-dom';
+import NotificationItem from './NotificationItem';
+import BulkSelect from '@redhat-cloud-services/frontend-components/BulkSelect';
+import axios from 'axios';
+import { Access } from '@redhat-cloud-services/rbac-client';
+import {
+  Stack,
+  StackItem,
+} from '@patternfly/react-core/dist/dynamic/layouts/Stack';
+import { getSevenDaysAgo } from '../UtcDate';
```
karelhala: It might be better to change this function to be `getSomeDaysAgo`? So we can get exact dates back?

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+  panelRef: React.Ref<unknown>;
+  isOrgAdmin: boolean;
+  toggleDrawer: () => void;
+  getUserPermissions: (
+    applicationName?: string,
+    disableCache?: boolean
+  ) => Promise<Access[]>;
+};
+
+const EmptyNotifications = ({
```
karelhala: Can we break this into a new file? Having multiple components in one file makes it harder to debug and read.

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+              limit: 50,
+              sort_by: 'read:asc',
+              startDate: getSevenDaysAgo(),
+            },
+          }
+        );
+        populateNotifications(data?.data || []);
+      } catch (error) {
+        console.error('Unable to get Notifications ', error);
+      }
```
karelhala: We have the JS client for notifications installed. Let's update the `notificationsApi` to include drawer endpoint and use it here instead of axios get.

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+          setFilterConfig(
+            response.data.map((bundle: Bundle) => ({
+              title: bundle.displayName,
+              value: bundle.name,
+            }))
+          );
+        }
+      } catch (error) {
+        console.error('Failed to fetch filter configuration:', error);
+      }
```
karelhala: Same as above, we do have the JS clients available, so we can add the bundles link to it and use it here as well.

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+        selectedNotifications.forEach((notification) =>
+          updateNotificationRead(notification.id, read)
+        );
+        setIsDropdownOpen(false);
+        updateAllNotificationsSelected(false);
+      })
+      .catch((e) => {
+        console.error('failed to update notification read status', e);
+      });
+  };
```
karelhala: Same as with fetch drawer data and facets. Let's use the JS client.

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+    onNotificationsDrawerClose();
+  };
+
+  const dropdownItems = [
+    <DropdownItem key="actions" description="Actions" />,
+    <DropdownItem
+      key="read selected"
+      onClick={() => {
+        onUpdateSelectedStatus(true);
+      }}
```
karelhala: Extra function body
```suggestion
      onClick={() => onUpdateSelectedStatus(true)}
```

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+      isDisabled={notifications.length === 0}
+    >
+      Mark selected as read
+    </DropdownItem>,
+    <DropdownItem
+      key="unread selected"
+      onClick={() => {
+        onUpdateSelectedStatus(false);
+      }}
```
karelhala: Extra function body
```suggestion
      onClick={() => onUpdateSelectedStatus(false)}
```

src/components/NotificationsDrawer/NotificationItem.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  const notificationDropdownItems = [
+    <DropdownItem key="read" onClick={onMarkAsRead}>{`Mark as ${
+      !notification.read ? 'read' : 'unread'
+    }`}</DropdownItem>,
+    <DropdownItem
+      key="manage-event"
+      onClick={() =>
+        onNavigateTo(
+          `/settings/notifications/configure-events?bundle=${notification.bundle}&tab=configuration`
```
karelhala, line 66: Since we are in the notifications application I think all links needs to be updated and the `/settings` should be removed. I would have to however have to check it locally what needs to remain and what can be removed.

src/components/NotificationsDrawer/DrawerPanel.tsx
```diff
@@ -0,0 +1,515 @@
...
+      isDisabled={notifications.length === 0}
+    >
+      Mark selected as unread
+    </DropdownItem>,
+    <Divider key="divider" />,
+    <DropdownItem key="quick links" description="Quick links" />,
+    <DropdownItem
+      key="notifications log"
+      onClick={() => onNavigateTo('/settings/notifications/notificationslog')}
```
karelhala: Same as with the `to` links. I think we can remove the `/settings` URL. But I'd have to check this locally to see what can remain and what can be removed.

src/components/NotificationsDrawer/DrawerUtils.ts
karelhala, line 1: Let's move this file to `__tests__` folder.

src/components/NotificationsDrawer/NotificationItem.tsx
```diff
@@ -0,0 +1,134 @@
...
+  notification,
+  onNavigateTo,
+  updateNotificationSelected,
+  updateNotificationRead,
+}) => {
+  const [isDropdownOpen, setIsDropdownOpen] = useState(false);
+
+  const onCheckboxToggle = () => {
+    updateNotificationSelected(notification.id, !notification.selected);
+  };
```
karelhala: Extra function body
```suggestion
  const onCheckboxToggle = () =>updateNotificationSelected(notification.id, !notification.selected);
```

src/components/NotificationsDrawer/NotificationItem.tsx
```diff
@@ -0,0 +1,134 @@
...
+        read_status: !notification.read,
+      })
+      .then(() => {
+        updateNotificationRead(notification.id, !notification.read);
+        setIsDropdownOpen(false);
+      })
+      .catch((e) => {
+        console.error('failed to update notification read status', e);
+      });
+  };
```
karelhala: Let's use the JS client for this function as well.

karelhala (commented):
Other than removing the cypress config, changing Atoms to Redux and utilizing JS clients just a few minor improvements.

config/webpack.cy.config.js
justinorringer: The PR did get big pretty quick haha. I threw these (imperfect) cypress stuff in a draft PR for you all to start with https://github.com/RedHatInsights/notifications-frontend/pull/667

src/DrawerSingleton.tsx
```diff
@@ -0,0 +1,109 @@
...
+        }
+      );
+      const dispatch = useDispatch();
+      dispatch(setNotificationsAction(data?.data || []));
+    } catch (error) {
+      console.error('Unable to get Notifications ', error);
+    }
+  };
+
+  public static DrawerSingletonInitializer = class {
```
Hyperkid123: We might even call the init function in the "get" if no instance exists and set some flags like "ready" so we have information for the eventual hooks.

src/DrawerSingleton.tsx
```diff
@@ -0,0 +1,109 @@
...
+        `/api/notifications/v1/notifications/drawer`,
+        {
+          params: {
+            limit: 50,
+            sort_by: 'read:asc',
+            startDate: getSevenDaysAgo(),
+          },
+        }
+      );
+      const dispatch = useDispatch();
```
Hyperkid123: You can't use hooks outside of react context. You will have to come up with something else. You need to tie "out of react scope" to "React scope". 😉 😉

src/components/NotificationsDrawer/NotificationItem.tsx
justinorringer: Messing around with these links for a minute, the /settings/notifications etc was not prepended. I'm thinking it's because the drawer doesn't go through the router?

Hyperkid123: @karelhala the pathname is not prepended. The drawer is not part of the notifications context. It is "its own app" if you will.

src/DrawerSingleton.tsx
justinorringer: I don't know where you got this lol, must have been an old commit

src/DrawerSingleton.tsx
justinorringer: Old code again

Hyperkid123 (approved):

PR #670, opened 2025-02-04
Change server status check to event endpoint

apinkert:
For [RHCLOUD-36131](https://issues.redhat.com/browse/RHCLOUD-36131). Since the `/api/notifications/v1.0/status` endpoint is going away, this PR replaces it with the `/api/notifications/v1.0/notifications/events` endpoint instead when doing a health check. Still trying to figure out why a unit test is failing, but wanted to get some eyes on it.

### DIFF
--- a/src/app/__tests__/App.test.tsx
+++ b/src/app/__tests__/App.test.tsx
@@ -61,13 +61,13 @@ jest.mock('@redhat-cloud-services/frontend-components/useChrome', () => {
 const mockMaintenance = (isUp: boolean) => {
   const response = isUp
     ? {
-        status: 'UP',
+        data: [],
+        links: {},
+        meta: {
+          count: 0,
+        },
       }
-    : {
-        start_time: '2021-06-11T13:09:31.213141',
-        end_time: '2021-06-11T18:09:31.213141',
-        status: 'MAINTENANCE',
-      };
+    : {};
 
   fetchMock.get('/api/notifications/v1.0/notifications/events', {
     status: 200,
--- a/src/app/__tests__/App.test.tsx
+++ b/src/app/__tests__/App.test.tsx
@@ -69,7 +69,7 @@ const mockMaintenance = (isUp: boolean) => {
         status: 'MAINTENANCE',
       };
 
-  fetchMock.get('/api/notifications/v1.0/status', {
+  fetchMock.get('/api/notifications/v1.0/notifications/events', {
     status: 200,
     body: response,
   });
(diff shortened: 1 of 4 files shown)


src/services/GetServerStatus.ts
```diff
@@ -5,11 +5,11 @@ import {
...
-import { Operations } from '../generated/OpenapiPrivate';
+import { Operations } from '../generated/OpenapiNotifications';
 import { toServer } from '../types/adapters/ServerAdapter';
 
 const adapter = validationResponseTransformer(
-  (payload: Operations.StatusResourceGetCurrentStatus.Payload) => {
+  (payload: Operations.EventResource$v1GetEvents.Payload) => {
     if (payload.status === 200) {
       return validatedResponse('ServerStatus', 200, toServer(), payload.errors);
     }
```
karelhala: You should also change the validator
```suggestion
    if (payload.status === 200) {
      return validatedResponse(
        'Events',
        200,
        {
          ...payload.value,
          data: payload.value.data.map(toNotificationEvent),
        },
        payload.errors
      );
    }
```

src/services/GetServerStatus.ts
```diff
@@ -5,11 +5,11 @@ import {
 } from 'openapi2typescript';
 import { useQuery } from 'react-fetching-library';
 
-import { Operations } from '../generated/OpenapiPrivate';
+import { Operations } from '../generated/OpenapiNotifications';
 import { toServer } from '../types/adapters/ServerAdapter';
```
karelhala: We no longer need the `toServer`
```suggestion
import { toNotificationEvent } from '../types/adapters/NotificationEventAdapter';
```

src/app/__tests__/App.test.tsx
```diff
@@ -69,7 +69,7 @@ const mockMaintenance = (isUp: boolean) => {
         status: 'MAINTENANCE',
```
karelhala: We also have to mock the response correctly
```JS
const response = isUp
    ? {
        data: [],
        links: {},
        meta: {
          count: 0,
        },
      }
    : {};
```

karelhala (commented):
The response needs to be adjusted to use the correct data. Alongside this we have to change the server config when using application to detect maintanance in `useApp` file:
```JS
if (serverStatus.payload?.type === 'Events') {
      if (serverStatus.payload?.status === 200) {
        setServer({
          status: ServerStatus.RUNNING,
        });
      } else {
        setServer({
          status: ServerStatus.MAINTENANCE,
          from: new Date(),
          to: new Date(),
        });
      }
    }
```

apinkert:
@karelhala I've updated responses and test mocks. Thanks for the comments!

karelhala (approved):

PR #676, opened 2025-02-21
Preselect events in edit integration wizard

CodyWMitchell:
### Description

Preselect previously added events in the edit integration wizard.

[RHCLOUD-36409](https://issues.redhat.com/browse/RHCLOUD-36409)

---

#### After:

---

### Checklist ☑️
- [X] PR only fixes one issue or story 
- [X] Change reviewed for extraneous code 
- [X] UI best practices adhered to 
- [X] Commits squashed and meaningfully named 
- [X] All PR checks pass locally (build, lint, test, E2E)

##
##

### DIFF
--- a/src/pages/Integrations/Create/CustomComponents/SelectableTable.tsx
+++ b/src/pages/Integrations/Create/CustomComponents/SelectableTable.tsx
@@ -15,6 +15,23 @@ import {
   EmptyStateIcon,
 } from '@patternfly/react-core';
 import CubesIcon from '@patternfly/react-icons/dist/dynamic/icons/cube-icon';
+import { getEndpoint } from '../../../../api/helpers/integrations/endpoints-helper';
+
+const BUNDLE_DEFAULTS = {
+  OpenShift: {},
+  'Red Hat Enterprise Linux': {},
+  Console: {},
+} as const;
+
+type EventBundle = keyof typeof BUNDLE_DEFAULTS;
+
+interface EventTypeMapping {
+  eventTypeDisplayName: string;
+  applicationDisplayName: string;
+  description: string;
+  id: string;
+  isSelected: boolean;
+}
 
 export interface TableRow {
   id: string;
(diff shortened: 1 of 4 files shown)


src/pages/Integrations/Create/CustomComponents/SelectableTable.tsx
```diff
@@ -73,6 +89,72 @@ const SelectableTable = (props) => {
...
+
+  type EventBundle = keyof typeof BUNDLE_DEFAULTS;
+
+  interface EventTypeMapping {
+    eventTypeDisplayName: string;
+    applicationDisplayName: string;
+    description: string;
+    id: string;
+    isSelected: boolean;
+  }
```
karelhala: Can you move these definitions out of the component please? It's much cleaner to define them before the component.

src/pages/Integrations/Create/eventTypesStep.ts
```diff
@@ -33,6 +33,7 @@ export const eventTypesStep = () => ({
       name: EVENT_TYPES_TABLE,
       label: 'Selected event types',
       bundleFieldName: 'product-family',
+      validate: [{ type: 'required' }],
```
karelhala, line 36: We can create integrations without event type. We'll probably want to implement a custom vlidator which would wait for the data to be loaded. Not sure how to do that, maybe load the data on initial wizard load?

@Hyperkid123 any suggestions? We want to wait until data are loaded for a step so when user performs edit the previous data are not lost.

CodyWMitchell: There might be a cleaner way to handle this, but the above `required` does not prevent them from submitting an integration without event types. It is being used here to wait until the initial data is loaded (which makes the `value` not `undefined`.

CodyWMitchell: Once the initial data loads, no event types would set the value to 
```
{
OpenShift: {},
'Red Hat Enterprise Linux': {},
Console: {},
}
```

No events chosen, but `required` is satisfied, enabling the Next button.

Hyperkid123: I'd say async validator. Or just hide the entire form until your data is loaded. Or is the data dependent on a specific step?

In that case, you might to create a custom spec template or a custom field. There are plenty of options.

karelhala: Ah, okay. If the required does not prevent from submitting empty event types we are fine with it.

CodyWMitchell: Would hiding the form on that step prevent the user from clicking **"Next"** in the Wizard? I'm not familiar with how this is typically handled in data-driven-forms.

CodyWMitchell: https://www.data-driven-forms.org/hooks/use-field-api#valid

I could create a validator that checks if the value is defined. The initial values are set once the request for the initial events completes, which allows them to proceed even if they don't modify their events.

Hyperkid123: Well if the value is [] or {} JS will resolve them as truthy. You could have a hidden field which is required. The user won't see it but the form will prevent you from clicking the next step until the hidden field is set. Maybe we can hop on a quick call so I can understand this case better.

karelhala (approved):

PR #686, opened 2025-03-06
Self initialize drawer global shared state.

Hyperkid123:
### Description
https://issues.redhat.com/browse/RHCLOUD-38614

Currently, the notification drawer state has to be explicitly initialized by some other entity (Currently Chrome). However, because the state is a "Singleton" and shared across modules that consume it, we can move the initialization check and call to the "subscribe" method.

This means, that invoking the subscription to the singleton state, will trigger the state initialization if it was not yet initialized. The state has to be initialized if some other modules are consuming it, so it is safe to keep it "pending" if no modules are trying to access it.

[RHCLOUDXXXX](https://issues.redhat.com/browse/RHCLOUDXXXX)

### Checklist ☑️
- [x] Change reviewed for extraneous code 
- [x] UI best practices adhered to 
- [x] Commits squashed and meaningfully named 
- [x] All PR checks pass locally (build, lint, test, E2E)

##
##

### DIFF
--- a/src/components/NotificationsDrawer/DrawerSingleton.tsx
+++ b/src/components/NotificationsDrawer/DrawerSingleton.tsx
@@ -1,4 +1,5 @@
-import { Access } from '@redhat-cloud-services/rbac-client';
+import { Access, AccessApi } from '@redhat-cloud-services/rbac-client';
+import axios from 'axios';
 
 import { getDateDaysAgo } from '../UtcDate';
 
@@ -11,6 +12,8 @@ import {
   NotificationData,
   NotificationDrawerState,
 } from '../../types/Drawer';
+
+const rbacApi = new AccessApi(undefined, '/api/rbac/v1', axios.create());
 
 interface Bundle {
   id: string;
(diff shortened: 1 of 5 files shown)


src/components/NotificationsDrawer/DrawerSingleton.tsx
```diff
@@ -26,14 +29,34 @@ const initialState: NotificationDrawerState = {
...
+    // Run the init procedure if the state is not ready for subscriber
+    if (!DrawerSingleton._state.initializing && !DrawerSingleton._state.ready) {
+      DrawerSingleton._state.initializing = true;
+      rbacApi
+        .getPrincipalAccess('notifications', undefined, undefined, 1000)
+        .then(({ data: { data } }) => {
+          DrawerSingleton.Instance.initialize(true, data);
+          DrawerSingleton._state.initializing = false;
+          DrawerSingleton._state.ready = true;
+          console.log(
```
justinorringer: Stray log

justinorringer (dismissed):
Makes a lot of sense! Didnt know we had a hook for that access call here :sob:

PR #732, opened 2025-06-17
Update integrations adapter with new spec values

apinkert:
Somewhere along the way, the OpenAPI spec was updated with these values but it was not updated in the adapter causing [RHCLOUD-38823](https://issues.redhat.com/browse/RHCLOUD-38823).

image


karelhala (dismissed):
Looking good, lint is failing

karelhala (approved):

PR #752, opened 2025-08-04
patch: use upcomming root widget CSS selectors

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-41601

Changes
Ensures that the root widget element class names match the future widget ID

### DIFF
--- a/src/components/Widgets/EventsWidget.scss
+++ b/src/components/Widgets/EventsWidget.scss
@@ -1,4 +1,4 @@
-.notifications-notificationsEvents {
+.notifications-notificationsEvents, [class*="notifications-notifications-./DashboardWidget"] {
   overflow-y: auto;
 }
 .pf-v5-c-empty-state {


src/components/Widgets/EventsWidget.scss
```diff
@@ -1,4 +1,4 @@
-.notifications-notificationsEvents {
+.notifications-notificationsEvents, [class*="notifications-notifications-./DashboardWidget"] {
```
aferd, line 1: just curious as to why you named this DashboardWidget instead of EventsWidget?

aferd (approved):

src/components/Widgets/EventsWidget.scss
Hyperkid123: That is based on what is exposed in your fec.config.js. The pattern of the selector is `scope-scope-module` where module is the identifier under which the component is exposed.

PR #762, opened 2025-08-27
feat(events-table): refactor events selection table to be DataView

karelhala:
### Description

Refactor events selection tables to use DataView and also remove old implementation of these tables.

Added new API version v2 to interact with events.

[RHCLOUD-37490](https://issues.redhat.com/browse/RHCLOUD-37490)

---

### Checklist ☑️
- [x] Change reviewed for extraneous code 
- [x] UI best practices adhered to 
- [x] Commits squashed and meaningfully named 
- [x] All PR checks pass locally (build, lint, test, E2E)

##
- [-] _(Optional) QE: Needs QE attention (OUIA changed, perceived impact to tests, no test coverage)_
- [-] _(Optional) QE: Has been mentioned_
- [-] _(Optional) UX: Needs UX attention (end user UX modified, missing designs)_
- [-] _(Optional) UX: Has been mentioned_
##


src/components/Integrations/EventTypes.tsx
```diff
@@ -0,0 +1,413 @@
...
+                      limit: newPerPage,
+                      offset: 0,
+                    },
+                    {}
+                  );
+                }}
+              />
+            }
+          />
+          <Table variant={'compact'} aria-label="Event types table">
```
InsaneZein, line 299: Could we use DataViewTable here instead of a regular Table since we're already using DataViewFilter and DataViewToolbar?

src/components/Integrations/EventTypes.tsx
```diff
@@ -0,0 +1,413 @@
...
+  setSelectedEvents,
+  currBundle,
+  applications,
+}) => {
+  const [loading, setLoading] = useState(true);
+  const [response, setResponse] = useState<{
+    meta?: { count: number };
+    data?: EventType[];
+  }>({});
+  const [expanded, SetExpanded] = useState<string[]>([]);
```
InsaneZein:
```suggestion
  const [expanded, setExpanded] = useState<string[]>([]);
```

src/components/Integrations/EventTypes.tsx
karelhala: Unfortunately not. I tried it and it was not working. The problem is that we have a mixture of collapsable rows with select boxes. I was unable to allow some of the rows to have collapsable items and checkboxes. If I used tree table variant the checkboxes were in all of children as well.

InsaneZein (dismissed):

PR #835, opened 2025-12-03
Convert events widget to dataview

apinkert:
For [RHCLOUD-37773](https://issues.redhat.com/browse/RHCLOUD-37773). The events widget was limited to 5 events. Converting it to DataView fixes this and we get pagination for free.

Before:
image

After:
image

### DIFF
--- a/src/components/Widgets/EventsWidget.tsx
+++ b/src/components/Widgets/EventsWidget.tsx
@@ -1,14 +1,5 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import {
-  Table,
-  TableVariant,
-  Tbody,
-  Td,
-  Th,
-  Thead,
-  Tr,
-} from '@patternfly/react-table';
-import {
   Button,
   ButtonVariant,
   EmptyState,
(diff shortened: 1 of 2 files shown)


src/components/Widgets/EventsWidget.tsx
```diff
@@ -49,63 +60,140 @@ const EmptyStateBellIcon: React.FunctionComponent = () => (
...
+    [perPage]
+  );
+
+  return (
+    <DataView
+      activeState={
+        loading ? 'loading' : notifications.length === 0 ? 'empty' : undefined
+      }
+    >
+      <Table aria-label="Events widget table" variant={TableVariant.compact}>
```
InsaneZein: I think we should be using DataViewTable for this and then we can pass loading and empty states as props for it.

apinkert: Ah good catch!

src/components/Widgets/EventsWidget.tsx
```diff
@@ -49,79 +55,156 @@ const EmptyStateBellIcon: React.FunctionComponent = () => (
...
-      {notifications.length === 0 ? (
-        <EmptyState
-          titleText={
-            <Title headingLevel="h4" size="lg">
-              {intl.formatMessage(messages.noFiredEventsTitle)}
-            </Title>
-          }
-          icon={EmptyStateBellIcon}
-          variant={EmptyStateVariant.full}
+  const handlePageChange = useCallback(
```
InsaneZein: not sure this or handlePerPageChange are necessary since we can just pass onSetPage and onPerPageSelect directly to the Pagination component.

InsaneZein (approved):
Ship it!

PR #841, opened 2025-12-12
RHCLOUD-41867 fix notificationsProvider

aferd:
### Description

notifications provider was broken during pf6 update. 

### Why this fixes the issue
The NotificationsProvider is rendered as a sibling (self-closing tag) rather than wrapping the child components. Looking at the package source, NotificationsProvider creates a React context and renders both a NotificationPortal and props.children. When used as a self-closing element, props.children is undefined, so only the portal is rendered but nothing is actually wrapped in the context.
The components using useNotifications() hook (like in AlertUtils.ts) won't have access to the context because they're not rendered inside the provider.
The fix should be to wrap the content with NotificationsProvider instead of using it as a sibling.

[RHCLOUD-41867](https://issues.redhat.com/browse/RHCLOUD-41867)

---

#### After:

---

### Checklist ☑️

##
##

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -83,18 +83,19 @@ const App: React.ComponentType = () => {
       }}
     >
       <RbacGroupContextProvider>
-        <NotificationsProvider />
-        <InsightsEnvDetector insights={insights} onEnvironment={staging}>
-          <RenderIfTrue>
-            <Switch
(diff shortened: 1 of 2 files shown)


jjaquish (approved):
Tested and working as expected locally. LGTM

PR #884, opened 2026-03-27
fix(event-log): add missing dropdown labels

Hyperkid123:
## Summary

- Fix empty timespan dropdown labels in Event Log and Notifications Log pages
- `SelectOption` components were self-closing with no children text, causing blank options in the dropdown
- Added label text as children to `SelectOption` in both `EventLogDateFilter.tsx` and `NotificationsLogDateFilter.tsx`

RHCLOUD-44667

## Verification

after

## Test plan

- [x] Existing tests pass (`EventLogFilter.test.tsx`)
- [x] Lint passes
- [x] Visual verification: dropdown now shows "Today", "Yesterday", "Last 7 days", "Last 14 days", "Custom"

### DIFF
--- a/src/components/Notifications/EventLog/EventLogDateFilter.tsx
+++ b/src/components/Notifications/EventLog/EventLogDateFilter.tsx
@@ -238,7 +238,9 @@ export const EventLogDateFilter: React.FunctionComponent<
   const options = React.useMemo(
     () =>
       Object.values(EventLogDateFilterValue).map((v) => (
-        <SelectOption key={v} value={new EventLogSelectObject(v)} />
+        <SelectOption key={v} value={new EventLogSelectObject(v)}>
+          {labels[v]}
+        </SelectOption>
       )),
     []
   );
--- a/src/components/Notifications/EventLog/__tests__/EventLogDateFilter.test.tsx
+++ b/src/components/Notifications/EventLog/__tests__/EventLogDateFilter.test.tsx
@@ -1,0 +1,80 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+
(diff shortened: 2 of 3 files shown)


Hyperkid123:
Hey our, please do not include the screenshot into the code changes, only to the PR description. Thank you!

Hyperkid123:
@Hyperkid123 Done! Removed the screenshot from the code changes in commit 026469a. Will include screenshots only in the PR description going forward. Thanks for the feedback!

Hyperkid123:
Can you please add unit tests for this change.

Hyperkid123:
@Hyperkid123 Added unit tests for both `EventLogDateFilter` and `NotificationsLogDateFilter` in commit a61f75a. Tests verify dropdown labels render correctly, option selection calls setValue, and custom date pickers appear only when Custom is selected (10 tests, all passing).

Hyperkid123:
Hello, bot. The code changes work and the bug is fixed. However the code seems to be almost identical, both in the source and for tests. Do you think you can deduplicate it? Also for a change like this, we do not need 4 commits. We want to keep the commit number small so it is easy to revert/cherry-pick changes. Can you amend that please.

Hyperkid123:
@Hyperkid123 Good feedback, addressed in commit 5df91b8:

- **Deduplicated tests**: Merged the two identical test files into a single parameterized test using `describe.each` that covers both `EventLogDateFilter` and `NotificationsLogDateFilter` (10 tests, 1 file).
- **Squashed commits**: Reduced from 4 commits to 1.

The source code duplication between the two filter components is pre-existing (not introduced by this PR) — the fix itself is just adding `{labels[v]}` as children to `SelectOption` in both files. Deduplicating the components themselves would be a larger refactoring beyond the scope of this bug fix.

InsaneZein (approved):

PR #975, opened 2026-06-15
feat(RBAC): Add v1 wildcard fallback for v2 orgs

aferd:
For v2 orgs, combine Kessel v2 permissions with v1 RBAC wildcard checks using OR logic. This ensures Org Admins and legacy roles with wildcard permissions (integrations:*:*, notifications:*:*) maintain access on v2 orgs.

Why: Kessel v2 does not currently support wildcard permission expansion. Without this fallback, users with wildcard permissions would lose access when their org is migrated to v2.

Temporary workaround: Should be removed once EITHER:
1. Kessel v2 supports wildcard permission expansion, OR
2. All wildcard permissions are migrated to explicit Kessel workspace relations

Related: insights-chrome PR #3362
Issue: RHCLOUD-48396

### Description

description text...

[RHCLOUDXXXX](https://issues.redhat.com/browse/RHCLOUDXXXX)

---

#### After:

---

### Checklist ☑️

##
##

### DIFF
--- a/src/app/useApp.ts
+++ b/src/app/useApp.ts
@@ -76,21 +76,50 @@ export const useApp = (): Partial<AppContext> => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  // Effect to fetch v1 RBAC when org is v1 (no need to wait for Kessel)
+  // Effect to fetch v1 RBAC for ALL orgs (v2 orgs need it for wildcard permission fallback)
+  // This is necessary because Kessel v2 does not support wildcard permission expansion,
+  // so we must check v1 wildcards (integrations:*:*, notifications:*:*) as fallback.
+  // See: https://github.com/RedHatInsights/insights-chrome/pull/3362
   useEffect(() => {
-    if (userLoaded && !isV2Org) {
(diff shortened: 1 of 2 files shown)


src/app/useApp.ts
```diff
@@ -76,21 +80,70 @@ export const useApp = (): Partial<AppContext> => {
...
   useEffect(() => {
-    if (userLoaded && !isV2Org) {
-      fetchRBAC(`${Config.notifications.subAppId},${Config.integrations.subAppId}`).then(setV1Rbac);
+    if (userLoaded) {
+      fetchRBAC(`${Config.notifications.subAppId},${Config.integrations.subAppId}`).then((rbac) => {
+        setV1Rbac(rbac);
+      });
     }
-  }, [userLoaded, isV2Org]);
+  }, [userLoaded, isOrgAdmin]);
```
jjaquish: Do we need isOrgAdmin here if we aren't checking it in the conditional?

aferd: ah, sure don't, you're right. good catch.

Bannermanj (approved):
lgtm

PR #1001, opened 2026-07-09
fix: Don't call Kessel API until workspace ID is available

riccardo-forina:
## Summary

- Splits `KesselRbacAccessProvider` into a gate that waits for the workspace ID and an inner `KesselPermissionChecker` that only mounts `useSelfAccessCheck` hooks once a valid workspace ID is available
- Eliminates the 400 validation errors (`object.resource_id: must be at least 1 characters`) caused by calling the Kessel API with an empty resource ID before the workspace had loaded
- Children still render during workspace loading with `isLoading=true` and all permissions `false` — no blank screen

### Root cause

The provider was creating a placeholder workspace with `id: ''` and immediately calling 7 `useSelfAccessCheck` hooks. The Kessel API rejected the empty resource ID with a 400. The hooks then re-ran with the correct ID once the workspace loaded, but the initial 7 failed requests contributed to the Akamai rate limit breach that caused the July 8 "Access Denied" incident.

### Before
```
T=0ms: Mount → workspace.id = '' → 7× useSelfAccessCheck → 7× POST checkself → 7× 400 error
T=200ms: Workspace loads → 7× useSelfAccessCheck re-runs → 7× POST checkself → 200 OK
```

### After
```
T=0ms: Mount → workspace loading → children render with isLoading=true, no API calls
T=200ms: Workspace loads → KesselPermissionChecker mounts → 7× useSelfAccessCheck → 7× POST checkself → 200 OK
```

14 Kessel calls → 7. Zero 400 errors.

## Test plan

- [ ] Load the notifications app as a V2 org admin — verify no 400 errors in network tab
- [ ] Verify notification drawer renders correctly
- [ ] Verify "Configure Events" appears in left navigation for org admins
- [ ] Verify permissions work correctly (can view events, edit integrations, etc.)

### DIFF
--- a/src/app/rbac/KesselRbacAccessProvider.tsx
+++ b/src/app/rbac/KesselRbacAccessProvider.tsx
@@ -1,142 +1,105 @@
 import React, { useEffect, useMemo, useState } from 'react';
 import { useSelfAccessCheck } from '@project-kessel/react-kessel-access-check';
+import type { SelfAccessCheckResourceWithRelation } from '@project-kessel/react-kessel-access-check';
 import { KesselRbacAccessContext, KesselRbacAccessContextValue } from './KesselRbacAccessContext';
-import { KESSEL_WORKSPACE_RELATIONS } from './kesselWorkspaceRelations';
+import {
+  KESSEL_WORKSPACE_RELATIONS,
+  KESSEL_WORKSPACE_RELATIONS_ORDERED,
+} from './kesselWorkspaceRelations';
 import { useDefaultWorkspace } from './hooks/useDefaultWorkspace';
 
+type AllowedByRelation = Partial<Record<string, boolean>>;
+
+function buildAllowedMap(
+  data: Array<{ allowed: boolean; relation: string }> | undefined
+): AllowedByRelation {
+  if (!data) {
+    return {};
+  }
+
+  const map: AllowedByRelation = {};
+  for (const item of data) {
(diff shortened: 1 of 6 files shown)


aferd (dismissed):
/lgtm

src/app/rbac/KesselRbacAccessProvider.tsx
```diff
@@ -1,142 +1,105 @@
...
+    })) as NonEmptyResources;
   }, [defaultWorkspaceId]);
 
-  // Permission checks - using v0.5.0 API which returns { loading, error, data }
-  const integrationsEndpointsEdit = useSelfAccessCheck({
-    relation: KESSEL_WORKSPACE_RELATIONS.INTEGRATIONS_ENDPOINTS_EDIT,
-    resource: workspace,
+  // The library requires NotEmptyArray but handles empty gracefully at runtime.
+  // When resources is undefined (no workspace), we pass [] to skip the API call.
+  const bulkResult = useSelfAccessCheck({
```
jjaquish, line 74: Do we need to track potential errors from useSelfAccessCheck here like for workspace errors above on lines 47-51?

jjaquish (approved):
I left one comment with a question, but otherwise this LGTM. I ran it locally and can see the fix working in the network logs. Otherwise the E2E tests seem to be in a similar state and some manual verification of features works as well.

PR #1011, opened 2026-07-13
feat: Add hidePagerDutySeverity prop to conditionally hide alert severity field

aferd:
## Summary
Read the `platform.integrations.pager-duty.hide-severity` Unleash feature flag directly in the IntegrationWizard component to conditionally hide the "Alert severity" field from the PagerDuty integration form.

This approach is **self-contained** - no changes needed in sources-ui. The wizard reads the flag directly using the existing `@unleash/proxy-client-react` integration.

## Changes
- **IntegrationWizard.tsx**: Read flag using `useFlag('platform.integrations.pager-duty.hide-severity')`
- **schema.ts**: Pass `hidePagerDutySeverity` parameter through to detailSteps
- **detailSteps.ts**: Conditionally include severity field based on flag using spread operator

## Behavior
- **Flag OFF (default)**: Alert severity field is visible and required
- **Flag ON**: Alert severity field is completely hidden from the form schema

## How It Works
```tsx
// IntegrationWizard.tsx
const hidePagerDutySeverity = useFlag('platform.integrations.pager-duty.hide-severity');

// detailSteps.ts - PagerDuty fields
...(hidePagerDutySeverity ? [] : [
  {
    component: componentTypes.SELECT,
    name: 'severity',
    label: 'Alert severity',
    // ... severity field definition
  },
]),
```

When the flag is `true`, the spread operator expands an empty array, effectively removing the field from the form schema. When `false`, the field is included normally.

## Testing
- [ ] Verify with flag OFF: Alert severity field is visible and required
- [ ] Verify with flag ON: Alert severity field is hidden from wizard
- [ ] Verify PagerDuty integration can be created without severity field when hidden
- [ ] Verify existing integrations continue to work
- [ ] Test in all entry points from sources-ui (no sources-ui changes needed):
  - [ ] IntegrationsDropdown
  - [ ] IntegrationsWidget  
  - [ ] Overview page

## Deployment
This is a standalone change. No coordination with sources-ui needed - the flag is read directly by this component.

**Steps:**
1. Merge this PR
2. Create Unleash flag `platform.integrations.pager-duty.hide-severity` (default: OFF)
3. Test in staging
4. Flip flag to ON on coordination day (with backend, docs, user communication)

## JIRA
RHCLOUD-48862

### DIFF
--- a/src/pages/Integrations/Create/IntegrationWizard.tsx
+++ b/src/pages/Integrations/Create/IntegrationWizard.tsx
@@ -142,6 +142,12 @@ export const IntegrationWizard: React.FunctionComponent<IntegrationWizardProps> 
             severity,
           }) => {
             const [type, sub_type] = intType?.split(':') || ['webhook'];
+            // When severity field is hidden during edit, preserve the original value
+            // Access severity safely - it exists on IntegrationPagerduty but not in the typed template
+            const templateSeverity =
+              template && 'severity' in template ? template.severity : undefined;
+            const severityValue =
+              hidePagerDutySeverity && isEdit && templateSeverity ? templateSeverity : severity;
             const data = {
               name,
               enabled: true,
@@ -153,7 +159,7 @@ export const IntegrationWizard: React.FunctionComponent<IntegrationWizardProps> 
                 url,
                 disable_ssl_verification: false,
                 secret_token,
-                severity,
+                severity: severityValue,
               },
               ...(isBehaviorGroupsEnabled && {
                 event_types: event_types
--- a/src/pages/Integrations/Create/IntegrationWizard.tsx
+++ b/src/pages/Integrations/Create/IntegrationWizard.tsx
@@ -103,6 +103,7 @@ export const IntegrationWizard: React.FunctionComponent<IntegrationWizardProps> 
   };
   const isBehaviorGroupsEnabled = useFlag('platform.integrations.behavior-groups-move');
   const isPagerDutyEnabled = useFlag('platform.integrations.pager-duty');
+  const hidePagerDutySeverity = useFlag('platform.integrations.pager-duty.hide-severity');
   const isEmailIntegrationEnabled = useFlag('platform.notifications.email.integration');
   const notifications = useNotification();
   const [wizardOpen, setWizardOpen] = React.useState<boolean>(isOpen);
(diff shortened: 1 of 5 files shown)


src/pages/Integrations/Create/detailSteps.ts
```diff
@@ -156,32 +161,37 @@ export const detailSteps = (isEdit: boolean, isBehaviorGroupsEnabled: boolean, i
...
-              value: 'Critical',
-            },
-          ],
-        },
+        ...(hidePagerDutySeverity
+          ? []
+          : [
+              {
+                component: componentTypes.SELECT,
+                name: 'severity',
```
jjaquish, line 170: Does this need validate like the other fields? Not sure if you consider that change outside the scope of this, but currently testing locally I'm able to go through the entire creation without setting severity and after finishing it just silently fails with a console error.

aferd: hmm, yeah, it probably should -- I will add the validate.

src/pages/Integrations/Create/IntegrationWizard.tsx
```diff
@@ -135,6 +142,12 @@ export const IntegrationWizard: React.FunctionComponent<IntegrationWizardProps>
             severity,
           }) => {
             const [type, sub_type] = intType?.split(':') || ['webhook'];
+            // When severity field is hidden during edit, preserve the original value
+            // Access severity safely - it exists on IntegrationPagerduty but not in the typed template
+            const templateSeverity =
+              template && 'severity' in template ? template.severity : undefined;
+            const severityValue =
+              hidePagerDutySeverity && isEdit && templateSeverity ? templateSeverity : severity;
```
jjaquish, line 150: If I understand correctly, if we're creating a new integration we won't hit isEdit and therefore severity will be undefined. Is that expected?

aferd: I went back and looked at this. so yes, that's expected. When hidePagerDutySeverity is true, the severity form field is removed from the schema entirely (in detailSteps.ts), so severity from the form values will be undefined. For new integrations with the flag on, we don't want to send a severity at all — the backend handles it. The conditional here is only needed for the edit case, where we want to preserve the existing severity value from the template so we don't accidentally clear it.

riccardo-forina (dismissed):
The labels in `detailSteps.ts` ('Alert severity', 'Info', 'Warning', 'Error', 'Critical' and the helperText) are hardcoded English strings. The `intl` object is already passed into this function — please use `intl.formatMessage()` for these fields. We're about to kick off the i18n initiative and these will need to be wrapped anyway.

Please fix before merge.

jjaquish:
@aferd We looked over this again in the review session today and were thinking it might be a good idea to have a simple test to check the feature flag works. What do you think?

jjaquish (approved):
LGTM

PR #1029, opened 2026-07-24
refactor: Consolidate v1/v2 nav permission coexistence in frontend.yaml

aferd:
## Summary
- Collapses duplicated v1/v2 Notifications nav groups into a single parent (no feature flag gate on the parent)
- Overview and Notification Preferences are now single records visible to all users (no duplication)
- Configure Events and Event Log keep dual v1/v2 records gated by `isITLess` + the appropriate permission method:
  - FedRAMP/Commercial (`isITLess(true)`): `loosePermissions` (v1 RBAC)
  - Regular cloud (`isITLess(false)`): `loosePermissionsKessel` (v2 Kessel)

## Dependencies
- **Requires** [RedHatInsights/insights-chrome#3623](https://github.com/RedHatInsights/insights-chrome/pull/3623) — exposes `isITLess` as a navigation visibility function in Chrome

## Context
Per Karel's finding (see [RHCLOUD-49549](https://redhat.atlassian.net/browse/RHCLOUD-49549) comments): FEO frontend.yaml supports v1/v2 coexistence via independent nav records. The `isITLess` check ensures the correct permission method is used per environment — FedRAMP environments don't have Kessel, so they must use v1 RBAC.

## Test plan
- [ ] Verify v1 org user sees: Overview, Configure Events (v1), Event Log (v1), Notification Preferences
- [ ] Verify v2 org user sees: Overview, Configure Events (v2), Event Log (v2), Notification Preferences
- [ ] Verify no duplicate nav entries appear for either user type
- [ ] Validate YAML against FEO schema

[RHCLOUD-49549]: https://redhat.atlassian.net/browse/RHCLOUD-49549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.rhcicd/frontend.yaml
+++ b/.rhcicd/frontend.yaml
@@ -45,68 +45,65 @@ objects:
         - segmentId: notifications-settings
           bundleId: settings
           navItems:
-            # V1 navigation (workspaces flag off)
             - title: Notifications
-              id: notifications-v1
+              id: notifications
               expandable: true
-              permissions:
-                - method: featureFlag
-                  args:
-                    - platform.rbac.workspaces
-                    - false
               routes:
-                - id: notificationOverview-v1
+                - id: notificationOverview
                   icon: PlaceholderIcon
                   title: Overview
                   href: '/settings/notifications'
+                # V1 permission check (workspaces flag off)
                 - id: configureEvents-v1
                   title: Configure Events
                   href: '/settings/notifications/configure-events'
                   permissions:
+                    - method: featureFlag
+                      args:
+                        - platform.rbac.workspaces
+                        - false
                     - method: loosePermissions
                       args:
                         - - notifications:*:*
                           - notifications:notifications:read
-                - id: eventLog-v1
-                  title: Event Log
-                  href: '/settings/notifications/eventlog'
-                  permissions:
-                    - method: loosePermissions
-                      args:
-                        - - notifications:*:*
-                          - notifications:events:read
-                - id: myUserPreferences-v1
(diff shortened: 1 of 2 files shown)


.rhcicd/frontend.yaml
```diff
@@ -45,68 +45,65 @@ objects:
...
                   title: Overview
                   href: '/settings/notifications'
+                # V1 permission check (workspaces flag off)
                 - id: configureEvents-v1
                   title: Configure Events
                   href: '/settings/notifications/configure-events'
                   permissions:
+                    - method: featureFlag
+                      args:
+                        - platform.rbac.workspaces
```
riccardo-forina: the `platform.rbac.workspaces` should not be used if not for gating different UX experiences like we do for IAM
In this case, we only need to parallel configs to the same route, with the different V1 and V2 access checks

riccardo-forina (changes requested):

.rhcicd/frontend.yaml
aferd: too fast, lol

aferd:
The test at integrations-ui.spec.ts:25 has test.fail() on line 29 — it was expected to fail due to a Chrome tab click bug (RHCLOUD-48620). Unrelated to this PR

aferd:
build failed due to e2e tests. We are expecting this to fail right now due to notif drawer not being available in stage/prod.

